PR #1, opened 2018-05-14
Masu skeleton

adberglund:
Application skeleton, some repo config files, and a status endpoint for Masu.


.github/ISSUE_TEMPLATE/Feature_request.md
```diff
@@ -0,0 +1,35 @@
...
+Ex. API, UI, Docs
+
+## Role
+
+
+## Assumptions
+-
+-
+
+## API Details
```
chambridge: Not important, but for this repo we wont likely have APIs defined like this or UI Details needed. Maybe just *Implementation Details*.

masu/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+logger = logging.getLogger(__name__)  # pylint: disable=invalid-name
+logger.addHandler(default_handler)
+
+
+def create_app(test_config=None):
+    """
+    App factory for Flask application.
+
+    Args:
+        test_config (dict): A mapping of configuratons used for testing
```
chambridge: spelling error.
configuratons --> configurations

masu/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    except OSError as e:
+        logger.warning(e)
+
+    # Establish database
+    # pylint: disable=invalid-name, unused-variable
+
+    db = SQLAlchemy(app)  # noqa: F841
+
+    # Routes
+    app.add_url_rule('/status/', view_func=StatusView.as_view('show_status'))
```
chambridge: Don't know enough about Flask. Will the url path be the same as koku (i.e. `/api/v1/status/`)? I think it will make it easy if we are consistent from a health check/ops perspective.

adberglund: Good point. The quickest way, given the very few endpoints likely here is just to prefix the route with `/api/v1` directly in the url_rule here. I believe we can also achieve this using Flask Blueprints, but I don't think they play nice with the class based view I wrote. Very happy to change that, thought. I went that route to mirror the koku API endpoint.

chambridge: I'm fine with just prefixing with `/api/v1` for now.

masu/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""App factory for Masu application."""
+import logging
+import os
+
+from flask import Flask
+from flask.logging import default_handler
+from flask_sqlalchemy import SQLAlchemy
```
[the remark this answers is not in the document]
adberglund: Okay. Let's find time this week to discuss.

My thinking is go as light as possible here. I know things tend to grow, but this realistically should have a status endpoint, and an endpoint to receive a message that new data is available--that endpoint will have to grow a little as we receive data from different sources. The rest is just pure Python processing. The only reason I see to use SQLAlchemy here is convenient session handling tied to requests. Otherwise I would use psycopg2. Once we're inserting data, we want to do so _en masse_, and we'll need to drop down to lower level than ORM for that to work.

PR #9, opened 2018-05-17
3 openshift

adberglund:
This should let us spin up masu in isolation (with its own copy of a DB) for local dev work, and put it into the project with koku already running

Ex. 

With openshift running, but no koku app

`make oc-dev-all` would spin up masu + a db.

If openshift is running and koku + db already exist

`make oc-create-masu` should put it in the `myproject` project where koku is deployed.


.coveragerc
```diff
@@ -4,6 +4,8 @@ branch = True
 source = masu
 omit =
   *config.py
+  tests/*
```
chambridge: Maybe:
```
*test/*
*/tests_*.py
```

Makefile
```diff
@@ -0,0 +1,101 @@
+PYTHON	= $(shell which python)
+
+TOPDIR  = $(shell pwd)
+PYDIR	= koku
+APIDOC = apidoc
```
chambridge: don't think we need this here.

Makefile
```diff
@@ -0,0 +1,101 @@
+PYTHON	= $(shell which python)
+
+TOPDIR  = $(shell pwd)
+PYDIR	= koku
```
chambridge: `masu`?

adberglund: 😴 @ the ☸️ again, clearly. Thanks for catching these.

PR #15, opened 2018-05-22
4 ingest s3 report

adberglund:
This is the simple file grab from S3. Added logic for processing and celery support is coming in the follow up stories.


Makefile
```diff
@@ -90,7 +90,7 @@ oc-rm-dev:
 
 oc-forward-ports:
 	-make oc-stop-forwarding-ports 2>/dev/null
-	oc port-forward $$(oc get pods -o jsonpath='{.items[*].metadata.name}' -l name=masu-pgsql) 15432:5432 >/dev/null 2>&1 &
+	oc port-forward $$(oc get pods -o jsonpath='{.items[*].metadata.name}' -l name=koku-pgsql) 15432:5432 >/dev/null 2>&1 &
```
blentz: Am I correct in interpreting this as masu won't be a standalone app, that you're expecting the dev environment to depend on the koku-pgsql DB?

masu/config.py
```diff
@@ -46,8 +46,11 @@ class Config(object):
     SQLALCHEMY_TRACK_MODIFICATIONS = False
 
     SECRET_KEY = os.getenv('MASU_SECRET_KEY')
-    if not SECRET_KEY:
+    if SECRET_KEY is None:
```
blentz: `if not SOME_VAR` is the correct, pythonic way of doing this check.

Under the covers, it calls the `.__nonzero__()` method to check truth, where doing `is None` *only* compares the SECRET_KEY reference to None.

masu/config.py
```diff
@@ -46,8 +46,11 @@ class Config(object):
...
         raise ValueError('No secret key set for Masu application')
 
     # Logging
     LOG_LEVEL = os.getenv('LOG_LEVEL', 'INFO')
+
+    # Data directory for processing incoming data
+    TMP_DIR = '/var/tmp/masu'
```
blentz: Based on the FHS, stuff in /var/tmp is temporary, but should persist across reboots where /tmp cannot be expected to persist across reboots.

IMO, using /tmp is fine, but I don't have a strong opinion. So, something to consider, but not a blocker for the PR.

masu/providers/aws/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+
+"""AWS Provider functions."""
```
blentz: We don't really need the legal stanza in `__init__.py` files unless there's actual code there. A simple `# noqa` is what's in most of our other `__init__.py` files.

tests/providers/aws/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+    'ap-southeast-2',
+    'ca-central-1',
+    'cn-north-1',
+    'cn-northwest-1',
+    'eu-central-1',
+    'eu-west-1',
+    'eu-west-2',
+    'eu-west-3',
+    'sa-east-1'
+)
```
blentz: Probably not necessary for testing purposes, but as an FYI, there is an API call to retrieve an authoritative list of regions - `ec2.describeRegions()`

tests/providers/aws/test_s3.py
```diff
@@ -0,0 +1,63 @@
...
+#
+
+"""Test the AWS S3 utility functions."""
+
+import os.path
+import random
+import tempfile
+from unittest.mock import ANY, Mock, patch, PropertyMock
+
+from faker import Faker
```
blentz: I'd like to see us using Moto for AWS tests - https://github.com/spulec/moto

I won't make you rewrite this right now, but I would like to see future PRs avoid reinventing this wheel. :smiley:

tox.ini
```diff
@@ -17,12 +17,12 @@ deps =
...
   codecov
 commands =
-  pipenv install --dev
+  pipenv install --dev --ignore-pipfile
   coverage run -m unittest
   coverage report --show-missing
 setenv =
   PYTHONPATH={toxinidir}
-
+  MASU_SECRET_KEY='603cbfcaec69ea552fc8993d673d16e7408c4e0398859d2bf95ac5a891e65d21'
```
blentz: How secret is this? Is this something that we're okay committing to a public git repo?

masu/config.py
adberglund: I think file download is a stand-in right now for dealing with a stream from S3. My reasoning behind using /var/tmp is that it *would* persist across reboots. This might not work exactly the same in container world (and there isn't currently a persistent volume set here, so I'm not sure we get anything out of this at the moment), but my thinking was we might crash in the middle of processing a file. When we come back to, the first check could be for a file in this directory. If one is there, process it. When we are done processing a downloaded file, we should delete it. It's bare bones disaster recovery.

PR #30, opened 2018-06-05
Initial Celery setup

blentz:


masu/processor/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+    # TODO: 1. Unzip and process each .csv.gz file in processing directory.
+    # TODO: 2. Insert any new data found in Step 1, skipping existing data.
+    pass
+
+
+@QUEUE.task(track_started=True)
+def test(args):
+    """Test task. Used only in testing."""
+    LOG.debug(args)
+    return mock.Mock(name=__name__, args=args)
```
dccurtis: I'm not sure if we should include test libraries in the production code.  Maybe we can do something like:
```
print("Start Task args: {}".format(args))
time.sleep(10)
print("End Task") instead?
```

masu/__init__.py
```diff
@@ -55,12 +57,18 @@ def create_app(test_config=None):
...
+        if e.errno != errno.EEXIST:
+            logger.warning(e)
 
     # Establish database
     # pylint: disable=invalid-name, unused-variable
     db = SQLAlchemy(app)  # noqa: F841
 
+    # Celery task queue
+    proc = Processor()
+    proc.run()
```
dccurtis: We might want to consider breaking out the run() from the initializer and have it kicked off by some other means.  That way we can have some more control in the test environment. For example: See if a task is ready (should return false), run a long task, check if task is done during and at the end.

masu/processor/tasks.py
```diff
@@ -0,0 +1,69 @@
...
+from celery.utils.log import get_task_logger
+
+from .celery.worker import QUEUE
+from .exceptions import MasuProcessingError
+
+
+LOG = get_task_logger(__name__)
+
+
+@QUEUE.task(autoretry_for=(MasuProcessingError,),
```
dccurtis: Will this QUEUE decorator cause make_celery to be ran each time?  It's been awhile since I've used celery but I remember there being just one instance of Celery() something like app = Celery() and you use @app.task.

tests/processor/test_tasks.py
```diff
@@ -0,0 +1,31 @@
...
+from unittest.mock import Mock
+
+from masu.processor.tasks import test
+from tests import MasuTestCase
+
+
+class TaskTest(MasuTestCase):
+    """Test Cases for the Processor functions."""
+
+    def test_test_task(self):
```
dccurtis: This is where I think we should be testing ready()

If we make the test task be some predefined long running task (10 seconds) then the test could be

```
ret = test('foo')
time.sleep(5)
self.assertIsFalse(ret.ready())
time.sleep(10)
self.assertIsTrue(ret.ready())
```

chambridge:
Do we need to update the README with any setup and content for local development?

masu/processor/tasks.py
blentz: I fully expect this method to get removed "Very Soon" once we have something more useful to do.

masu/__init__.py
blentz: Agree, but I'm expecting that to come with additional development beyond this very rudimentary scaffolding. Most of that requires spec'ing out interactions we haven't yet defined.

blentz:
@chambridge the dev setup is mostly handled by the openshift template (and also the masu README is totally empty right now. :sweat_smile: )  But yes, I agree with you - we probably need to start fleshing out those docs. I think I'll work that up in a separate PR, if that works for you.

masu/processor/celery/worker.py
```diff
@@ -0,0 +1,50 @@
...
+#
+"""Celery worker module."""
+
+import logging
+
+from celery import Celery
+from flask import Flask
+
+
+def make_celery(app):
```
adberglund: I'm not following why we have a processor and this celery app factory? I think this factory function is all we need? What am I missing?

masu/processor/celery/worker.py
```diff
@@ -0,0 +1,50 @@
...
+            """Call task with context."""
+            with app.app_context():
+                return self.run(*args, **kwargs)
+
+    celery.Task = ContextTask
+    return celery
+
+
+LOG = logging.getLogger(__name__)
+FLASK = Flask(__name__, instance_relative_config=True)
```
adberglund: You would want to create an instance of our app here, which is made using the create_app app factory, I believe? I was thinking we would create the celery app inside our app factory so that it would always be available in our app.

masu/__init__.py
adberglund: I think we should be using the celery app factory you created in worker.py here. And it really just becomes our plugin point to celery.

```
celery = make_celery(app)
```

masu/config.py
```diff
@@ -45,6 +45,10 @@ class Config(object):
 
     SQLALCHEMY_TRACK_MODIFICATIONS = False
 
+    # AMQP Message Broker
+    RABBITMQ_HOST = os.getenv('RABBITMQ_HOST', 'localhost')
+    RABBITMQ_PORT = os.getenv('RABBITMQ_PORT', '5672')
```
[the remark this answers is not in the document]
blentz: No user or password is created. Using guest access for now, mostly because everything is on a private internal network that's inaccessible outside the pod.

masu/processor/celery/worker.py
blentz: `make_celery()` is straight out of the Celery docs.

I created the `Processor` class for organization purposes. I haven't yet decided which idioms make sense to use, but I'm starting to lean towards more object-oriented. At some point, I think the factory function will move. 

But, the short answer is that the discrepancy is a product of me iterating through the tutorial docs until I'm comfortable enough with what they describe to start thinking about how I want to organize the structures.

masu/processor/celery/worker.py
blentz: I was having trouble seeing how to accomplish that in a satisfying way.

Also, `worker` is what is called by the openshift template to spawn the workers, which is a secondary entry point from the Flask app itself.

masu/__init__.py
blentz: I think the main issue is what comes next?  We have to call [task].delay() somewhere. So, where does that get called if `get_app` is calling `make_celery` ?

The options I can think of are:

1. call tasks in `get_app`, which is the Big Ball Of Mud design pattern, and is not ideal.
2. create an object and pass in the celery object obtained from `make_celery` into our task-running object, which is acceptable but begs the question of why doesn't that object's constructor just use the factory directly?
3. Insert an object between `get_app` and the celery factory to handle the interface between Flask and our processing tasks. Maybe we absorb the factory method into our object, and maybe we don't. This is where I stopped short, to spend time looking at other parts of the design before committing to pushing the factory into `Processor` entirely.

adberglund: We call [task].delay() wherever/whenever we want to put a task in the queue. In the SNS example, that would be done in the view that handles the endpoint. In the polling case if we switch to using the scheduler, then we don't call `.delay()` anywhere.

masu/celery/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+
+from unittest import mock
+
+from celery.utils.log import get_task_logger
+
+from masu import create_app
+from masu.celery import create_celery
+from masu.processor.exceptions import MasuProcessingError, MasuProviderError
+
+MASU = create_app()
```
adberglund: I think I probably led you astray of the @shared_task the other day, but I think we can ultimately use them here so we don't have to create an app or celery instance. the autodiscovery process in the app factory will then find all of these and associate them with the celery instance we make there. We can experiment with that later when we build these tasks out further.

PR #31, opened 2018-06-05
Celery poc

adberglund:
This is **not** intended to be merged in, but it was the quickest way to show what I am thinking. I'll kill the branch after taking a look at it.


masu/celery_worker.py
```diff
@@ -0,0 +1,6 @@
+from masu import create_app
+from masu.celery import make_celery
+
+app = create_app()
+app.app_context().push()
+celery = make_celery(app)
```
adberglund: This is callable as `celery worker -A masu.celery_worker.celery`

PR #34, opened 2018-06-07
Running tox will create the database for tests

adberglund:
This relies on environment variables being set as in the `.env.example`.

🤞 that travis will actually run with this setup.

I did not give much thought to file names or where things live. Please advice.

The thinking is that running either the make command or tox will get the database ready for tests to be run. We'll have to have setup and teardown methods on our tests to create and delete what we put in the database since we don't get it for free like in Django

Steps to create the SQL dump file:

1) Create a fresh database.
2) Run migrations in koku
3) Create a customer named `Test Customer`
4) pg_dump -d koku -h localhost -p 15432 -U kokuadmin > test.sql


Makefile
```diff
@@ -112,3 +113,8 @@ oc-stop-forwarding-ports:
 oc-serve: oc-forward-ports
 	export FLASK_APP=masu; export FLASK_ENV=development; flask run
 	make oc-stop-forwarding-ports
+
+oc-run-migrations: oc-forward-ports
+	sleep 1
+	$(TOPDIR)/tests/create_db.sh
```
dccurtis: Can we add this script to run-migrations so it can be generated when running locally?

PR #36, opened 2018-06-07
Database abstraction layer to pull in Provider data into Masu.

dccurtis:
Masu and Koku share a common database to relay information such as customer and provider data.  This data is relevant to Masu because it provides the customer list and cost usage report access credentials to the cloud provider.

**Overview of the change**
* Establish a database abstraction to isolate all sqlalchemy commands and provide a more convenient object-oriented architecture for masu-core to consume.
* The testing at this point is simply verifying that we can read values from the pre-loaded test database.


.env.example
```diff
@@ -1,6 +1,7 @@
 DATABASE_ENGINE=postgresql
 DATABASE_NAME=koku
+#DATABASE_NAME=test
 DATABASE_HOST=localhost
 DATABASE_PORT=15432
 DATABASE_USER=kokuadmin
-DATABASE_PASSWORD=''
+DATABASE_PASSWORD=kokuadmin
```
blentz: Why are we changing values in the example file? It's an example, so it shouldn't be getting used for anything.

masu/database/provider_billing_source_db_accessor.py
```diff
@@ -0,0 +1,45 @@
...
+        obj = self.get_session().query(self._billing_source).filter_by(id=self._billing_source_id)
+        return obj
+
+    def get_uuid(self):
+        """Return the provider uuid."""
+        obj = self._get_db_obj_query().first()
+        return obj.uuid
+
+    def get_bucket(self):
+        """Return the provider resource name."""
```
blentz: Is it returning a bucket name (e.g. "my-bucket-name") or an ARN (e.g. "aws::s3::whatever/my-bucket-name")?

Docstring should include an example of the return value.

.env.example
dccurtis: Yes.  I'll revert this.  Didn't mean to check this in.

masu/database/provider_billing_source_db_accessor.py
dccurtis: Yes, I'll update the docstrings.  Good call.

dccurtis:
Thanks for the comments @blentz.  I agree with you on the unknown return value format.  I was basically trying to keep the methods to be in sync with what koku and the db have.  For better or worse, I thought it might make debugging with looking at the database easier since the names would al match.

I'm open to changing the getter names to be more descriptive but for now I'll update the docstrings to make the class more reader friendly.

.env.example
dccurtis: I'll revert it.  The DATABASE_NAME however, might be helpful to keep since to run the unit tests the database name must be test.

blentz: You should not rely on the example file for testing.

masu/database/provider_billing_source_db_accessor.py
adberglund: Perhaps we can agree on a docstring style so we all know what to expect. #codingstandards

.env.example
adberglund: @dccurtis I think an improvement on these lines that we could use is to set appropriate env variables in the tox.ini. I just kind of lazily pulled in everything through `passenv` which I was thinking of the travis use case primarily for. The next section is `setenv` where we could really use an override `DATABASE_NAME=test` could go there quite nicely. Likewise if I'm missing any others that need to get overridden for tests.

dccurtis:
I know some of the naming of things like 'Provider Resource Name" being "Authentication" isn't great.  But I wanted to keep the naming consistent between koku and the database for this layer.  We could consider changing the names overall or I could make simple wrappers with a more descriptive name to be more convenient within Masu.

masu/database/auth_db_accessor.py
```diff
@@ -0,0 +1,61 @@
...
+        obj = self.get_session().query(self._auth_group).filter_by(id=self._auth_id)
+        return obj
+
+    def get_name(self):
+        """
+        Return the name for the authentication group.
+
+        Args:
+            None
+        Returns:
```
adberglund: Since we are talking convention... can we put a blank line between each section and I don't think we need double quotes on the description of the return type. Definitely not holding this back on that, just stating my preference. Thoughts?

```
Args:
    arg

Returns:
    return_type: description
        example
```

dccurtis: Agreed.  I'll touch things up in a later change.

PR #40, opened 2018-06-12
add ReportDownloader

blentz:


masu/processor/providers/aws/report_downloader.py
```diff
@@ -0,0 +1,208 @@
...
+            (String): The path and file name of the saved file
+
+        """
+        s3_filename = key.split('/')[-1]
+        file_path = f'{DATA_DIR}/{self.customer_schema}/aws'
+        file_name = f'{file_path}/{s3_filename}'
+        # Make sure the data directory exists
+        os.makedirs(file_path, exist_ok=True)
+
+        etag = self.report_stats.get_etag(key)
```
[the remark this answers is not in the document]
blentz: Key is the path+name of the object in S3.  (e.g.  `/my-report-path/some-dir/foo.csv`)

masu/processor/report_downloader.py
```diff
@@ -0,0 +1,39 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Report Downloader."""
+import logging
+from tempfile import mkdtemp
+
+
+# pylint: disable=too-few-public-methods
+class ReportDownloader():
```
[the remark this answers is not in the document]
blentz: Agreed. I stubbed it out this way with the expectation that we'd sub-class for various cloud providers (e.g. `AzureReportDownloader`, `GCPReportDownloader`).  There's likely to be common functionality that we can pull into a base class.

I'm agnostic on whether it should be a base class or a super-class.

masu/processor/providers/aws/report_downloader.py
```diff
@@ -0,0 +1,208 @@
...
+        for s3obj in bucket.objects.all():
+            files.append(self.download_file(s3obj.key))
+        return files
+
+    def download_current_report(self):
+        """
+        Read CUR manifest, download current report files.
+
+        Returns:
+            (String): The path and file name of the saved file
```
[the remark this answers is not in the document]
blentz: I believe I marked where Lists are returned instead of Strings.  `download_file()` returns a single string because it downloads a single file. `download_current_report()` returns a list because the report can span multiple files.

masu/processor/providers/aws/report_downloader.py
```diff
@@ -0,0 +1,208 @@
...
+        super().__init__(**kwargs)
+
+        self.provider = kwargs.get('provider')
+        self.customer = kwargs.get('customer')
+        self.report_stats = kwargs.get('report_stats')
+
+        self.customer_schema = self.customer.get_schema_name()
+
+        self.logger.debug('Connecting to AWS...')
+        session = get_assume_role_session(
```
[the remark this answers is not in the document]
blentz: +1 agree. I haven't handled those exceptions yet, primarily because I hadn't fully decided _where_ to handle them yet.

I was debating whether to catch the exceptions at the caller and re-raise when needed versus catching the exception later on. So, I left it alone until I needed to become more opinionated on where to catch the thrown exceptions.

masu/processor/providers/aws/report_downloader.py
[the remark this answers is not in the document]
dccurtis: I was looking at what download_report does.  It looks like it gets the manifest for the specified date and then downloads all of the files enumerated in the manifest.  Thats why I was under the impression that download_current_report could potentially download multiple files.

masu/processor/providers/aws/report_downloader.py
```diff
@@ -0,0 +1,209 @@
...
+        Returns:
+            (List) List of filenames downloaded.
+
+        """
+        s3_resource = boto3.resource('s3')
+        bucket = s3_resource.Bucket(self.bucket)
+        files = []
+        for s3obj in bucket.objects.all():
+            files.append(self.download_file(s3obj.key))
+        return files
```
adberglund: Downloading an entire bucket could be a lot of data. What's our use case?

masu/processor/providers/aws/report_downloader.py
```diff
@@ -0,0 +1,209 @@
...
+        os.makedirs(file_path, exist_ok=True)
+
+        etag = self.report_stats.get_etag(key)
+        s3_file = self.s3.get_object(Bucket=self.bucket, Key=key)
+        if s3_file.get('etag') != etag:
+            self.logger.info('Downloading %s to %s', s3_filename, file_name)
+            self.s3.download_file(self.bucket, key, file_name)
+            self.report_stats.update(report_name=key,
+                                     etag=s3_file.get('etag'),
+                                     last_completed_datetime=datetime.now())
```
adberglund: I might be wrong in the intention, but I was thinking that the last_completed_datetime would be reserved for when the report processing was completed, not just the etag updated. I would read that as the report with that etag has already been processed.

tests/processor/providers/aws/test_report_downloader.py
```diff
@@ -0,0 +1,279 @@
...
+        self.mock_provider = Mock(**{
+            'get_provider_resource_name.return_value': self.fake.word().lower(),
+            'get_report_name.return_value': self.fake_report_name})
+
+        self.mock_report_stats = Mock()
+
+        self.report_downloader = AWSReportDownloader(**{'customer': self.mock_customer,
+                                                        'provider': self.mock_provider,
+                                                        'report_stats': self.mock_report_stats})
+        self.assertIsInstance(self.report_downloader, AWSReportDownloader)
```
adberglund: This assertion probably isn't needed. You just make an object of that type.

masu/processor/providers/aws/report_downloader.py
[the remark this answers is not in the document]
dccurtis: I think it could make sense to catch the exceptions here, log it, and then raise an AWSReportDownloader exception.  That way the caller of AWSReportDownloader will know at instantiation time that the object is not usable and it could handle things.

This will be the orchestrator that will have to deal with a useless downloader and it would have to signal to the service admin some how some explanation of whats wrong (invalid credentials, networking problems, etc).

The orchestrator integration will come next but for now I think it's good to have the downloader absorb and interpret the boto3 exception and raise it's own.

masu/processor/report_downloader.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Download cost reports from a provider.
+
+    Base object class for downloading cost reports from a cloud provider.
+    """
+
+    def __init__(self, **kwargs):
```
adberglund: Again here we have a single key-wordable argument right now. I think we should probably be explicit in the init. The only keyword argument that is currently handled below is `download_path`.  Let's just say `__init__(self, download_path='')`

Then `self.download_path = download_path if download_path else mkdtemp(prefix='masu')`

masu/processor/providers/aws/report_downloader.py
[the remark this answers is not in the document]
dccurtis: By the way I'm good with leaving this exception work to be done when we integrate with the orchestrator.  Just leaving some thoughts here.

tests/processor/test_report_downloader.py
```diff
@@ -0,0 +1,49 @@
...
+        self.assertIsInstance(downloader, ReportDownloader)
+        self.assertIsNotNone(downloader.download_path)
+        self.assertTrue(os.path.exists(downloader.download_path))
+
+    def test_report_downloader(self):
+        dl_path = '/{}/{}/{}'.format(self.fake.word().lower(),
+                                     self.fake.word().lower(),
+                                     self.fake.word().lower())
+        downloader = ReportDownloader(download_path=dl_path)
+        self.assertIsInstance(downloader, ReportDownloader)
```
adberglund: Same thing with these isinstance checks. the variable was just instantiated as that object type.

tests/providers/aws/test_s3.py
```diff
@@ -1,63 +0,0 @@
...
-        mock_os.return_value = None
-
-        with tempfile.TemporaryDirectory() as tmp_dir:
-            mock_bucket.download_file.side_effect = save_file(tmp_dir, key)
-
-            result = s3.get_file_from_s3(schema, region, bucket_name, key)
-
-            self.assertEqual(result, expected)
-            # See if the file exists in the directory
-            self.assertTrue(os.path.isfile(f'{tmp_dir}/{key}'))
```
adberglund: How are we testing that we are actually storing the file we download in the new tests above?

masu/processor/providers/aws/report_downloader.py
blentz: It's mostly just for development purposes. I'm uncertain if the on-boarding case is better to read-through the manifest to figure out what to download, or to simply brute-force it. It was trivial to provide both cases to allow us to experiment with it.

masu/processor/providers/aws/report_downloader.py
blentz: Okay, that's fine. 

I wanted a way to signal that we've actually downloaded new files versus just returned the list of already-cached files. I can have the Downloader return a more complex object that includes a `{changed: True}` dict. Or maybe there's a third option? Thoughts?

tests/providers/aws/test_s3.py
blentz: We don't need to duplicate testing that the boto API works. That's boto's job to test that.

PR #44, opened 2018-06-15
Introducing CUR Statistics DB Accessor

dccurtis:
**Overview**
This change is in support of #108. When processing cost usage reports we want to keep various statistics including:

- Last processing start time
- Last completed time
- File cursor position
- S3 object etag

These parameters will be used to optimize our file processing resources.

**What's changed**

- Database Accessor to be able to read/write the database owned by Koku
- Basic integration with the CUR downloader to be able to store the file's etag and to block unnecessary downloads

**Remaining related work**
- When CUR processing is added we will have to log the last started/completed and cursor positions.
- Any database access patterns that are done as part of CUR processing may need to be adopted to this change.
-  Possibly add some relationship to the report it self (currently we are keeping track of individual .csv files that are part of a report)
- When the Processor Orchestrator is invoking the downloader we might want to pull ProviderDBAccessor out of the downloader and have the downloader take arguments for the S3 bucket and RoleARN since we are going to be iterating over something that looks similar to ProviderDBAccessor to get these values.


masu/processor/providers/aws/report_downloader.py
```diff
@@ -40,42 +41,40 @@ class AWSReportDownloader(ReportDownloader):
...
-    def __init__(self, customer, provider, report_stats, **kwargs):
+    def __init__(self, provider, report_name, **kwargs):
         """
         Constructor.
 
         Args:
-            customer     (CustomerDBAccessor) the customer accessor
-            provider     (ProviderAuthDBAccessor) the provider accessor
-            report_stats (ReportStatsDBAccessor) the report stats accessor
+            provider     (ProviderDBAccessor) the provider accessor
```
chambridge: Missing `report_name` in the docstring.

tests/providers/test_report_stats_db_accessor.py
```diff
@@ -0,0 +1,138 @@
...
+from datetime import datetime
+
+from masu.database.report_stats_db_accessor import ReportStatsDBAccessor
+from tests import MasuTestCase
+
+class ReportStatsDBAccessorTest(MasuTestCase):
+    """Test Cases for the ReportStatsDBAccessor object."""
+
+    def test_initializer(self):
+        '''Test Initializer'''
```
chambridge: quote consistency.
``'''Test Initializer'''`` --> ``"""Test Initializer"""``

masu/database/report_stats_db_accessor.py
```diff
@@ -0,0 +1,215 @@
...
+
+        Args:
+            (Integer): last byte offset for the given CUR file.
+        Returns:
+            None
+
+        """
+        obj = self._get_db_obj_query()
+        new_atribute = {'cursor_position': new_position}
+        obj.update(new_atribute)
```
adberglund: Looking at http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.update see the warning. We are losing the benefit of the ORM by using `query.update()`

If we go directly with something like 

```
obj = self._get_db_obj_query().first()
obj.cursor_position = new_position
```

That puts the session in a "dirty" state, and the session tracks changes for us, and we're back in the ORM's good graces. 

Also should this just call the update method like in log_last_started_datetime below?

masu/database/report_stats_db_accessor.py
```diff
@@ -0,0 +1,215 @@
...
+        if cursor_position:
+            update_dict['cursor_position'] = cursor_position
+        if last_completed_datetime:
+            update_dict['last_completed_datetime'] = last_completed_datetime
+        if last_started_datetime:
+            update_dict['last_started_datetime'] = last_started_datetime
+        if etag:
+            update_dict['etag'] = etag
+
+        obj_to_update.update(update_dict)
```
adberglund: Same comment as above, instead of using the query's update method, I think we should just modify the object's attributes directly and then you can just commit the session to save. 

```
if cursor_position:
    obj.cursor_position = cursor_position
...

self.commit()
```

masu/processor/providers/aws/report_downloader.py
dccurtis: Good catch.  Updated.

masu/database/report_stats_db_accessor.py
```diff
@@ -0,0 +1,215 @@
...
+    def get_cursor_position(self):
+        """
+        Return current cursor position for processing CUR.
+
+        Args:
+            None
+        Returns:
+            (Integer): last byte offset for the given CUR file.
+        """
+        obj = self._get_db_obj_query().first()
```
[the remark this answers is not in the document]
dccurtis: Yeah, thats a good idea.  Each Accessor object mainly references a single object, we might as well do that in the initializer.  As you mentioned in your other comment, we should add this to a list of things we want to tweak for interacting with sqlalchemy.  I'm sure there are many other things we want to change across the board after we get full processing completed.  I'm good with doing these types of changes after we get the basic pipeline components in place.

masu/database/report_stats_db_accessor.py
dccurtis: Yes agree with this.  I can try to make this change here just so we know it works and then adopt this pattern across the other accessors (if there are anyway).

As for calling update internally.  Yeah its inconsistent now.  I'll experiment with this and try to make the change to move away from using update.

PR #45, opened 2018-06-15
Create CUR Account interface

dccurtis:
**Objective**
Masu gets all of the CUR download credentials from Koku.  It does this whenever a provider is added by the user.  This interface will gather all of the accounts and provide an opaque interface for masu to consume.

**Implementation Details**
The interface will supply all of the account details needed to download a CUR from the cloud provider.  A layer of indirection is used to give us the flexibility to source account details from other places aside from the Koku database.  The most obvious alternative that comes to mind is a REST API call.

- There is a database layer which will query all Providers
- The provider details are gathered and generalized so that they can conform to a pre-defined CUR Account interface
- Masu will use the CUR Account interface to iterate over all accounts and get the needed information for CUR download.

Note: AccountsAccessorTest demonstrates how this will be used by the Masu Orchestrator

```
accounts = AccountsAccessor().get_accounts()

for account in accounts:
    access_cred = account.get_access_credential()    # RoleARN
    cur_source = account. get_billing_source()       # S3 Bucket
    customer_name = account.get_customer()           # Name of customer setup in Koku
```

**Testing**
Unit test is used to verify this change.  We can externally verify the functionality once it is used by the CUR Downloader and Orchestrator.


masu/external/accounts_accessor.py
```diff
@@ -0,0 +1,70 @@
...
+    """Cost Usage Report Accounts error."""
+
+    pass
+
+
+# pylint: disable=too-few-public-methods
+class AccountsAccessor:
+    """Interface for masu to use to get CUR accounts."""
+
+    def __init__(self, source='db'):
```
chambridge: maybe update the param to `source_type`; seem strange to assign source to `self.source_type` and then assign `self. source` to something else.

tests/database/test_provider_collector.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class ProviderQueryTest(MasuTestCase):
+    """Test Cases for the ProviderDBAccessor object."""
+
+    def setUp(self):
+        pass
+
+    def test_initializer(self):
+        '''Test Initializer'''
```
chambridge: Nit, but can we be consistent with quotes?
``'''Test Initializer''' `` --> ``"""Test Initializer"""``

tests/external/db/test_cur_accounts_db.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class CURAccountsDBTest(MasuTestCase):
+    """Test Cases for the CURAccountsDB object."""
+
+    def setUp(self):
+        pass
+
+    def test_get_accounts_from_source(self):
+        '''Test to get all accounts'''
```
chambridge: Nit, but can we be consistent with quotes?
``'''Test to get all accounts'''`` --> ``"""Test to get all accounts"""``

tests/external/test_accounts_accessor.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class AccountsAccessorTest(MasuTestCase):
+    """Test Cases for the AccountsAccessor object."""
+
+    def setUp(self):
+        pass
+
+    def test_get_accounts(self):
+        '''Test to get_access_credential'''
```
chambridge: Nit, but can we be consistent with quotes?
``'''Test to get_access_credential'''`` --> ``"""Test to get_access_credential"""``

tests/external/test_accounts_accessor.py
```diff
@@ -0,0 +1,49 @@
...
+
+        self.assertIsInstance(account_objects.pop(), CostUsageReportAccount)
+
+        for account in account_objects:
+            self.assertEqual(account.get_access_credential(), 'arn:aws:iam::111111111111:role/CostManagement')
+            self.assertEqual(account.get_billing_source(), 'test-bucket')
+            self.assertEqual(account.get_customer(), 'Test Customer')
+
+    def test_invalid_source_specification(self):
+        '''Test that error is thrown with invalid account source.'''
```
chambridge: Nit, but can we be consistent with quotes?
``'''Test that error is thrown with invalid account source.'''`` --> ``"""Test that error is thrown with invalid account source."""``

tests/external/test_cur_account.py
```diff
@@ -0,0 +1,47 @@
...
+    def test_get_access_credentials(self):
+        '''Test to get_access_credential'''
+        self.assertEqual(self.account.get_access_credential(), self.auth_string)
+
+    def test_get_billing_source(self):
+        '''Test to get_billing_source'''
+        self.assertEqual(self.account.get_billing_source(), self.billing_source)
+
+    def test_get_customer_name(self):
+        '''Test to get_customer_name'''
```
chambridge: Same quote comments.

tests/database/test_auth_db_accessor.py
```diff
@@ -28,7 +28,7 @@ def setUp(self):
         pass
 
     def test_initializer(self):
-        '''Test Initializer'''
+        """Test Initializer"""
```
adberglund: 🙇

masu/external/cost_usage_report_account.py
```diff
@@ -0,0 +1,73 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Cost Of Usage Report (CUR) object to be used by Masu Orchestrator/Processor to download CURs."""
+
+
+class CostUsageReportAccount:
```
adberglund: I think I get it, but say once we have a 2nd provider type, that shouldn't matter much here, right? It'll just return an account that will give us enough info to know which provider type to process? Or is this class supposed to be specific to AWS?

dccurtis: Thats right, a second provider will work the same way.  The idea of CostUsageReportAccount is to abstract away all of the details of AWS or whatever.  Once we support multiple providers things may have to change slightly to accommodate the other provider but the intention is to be provider agnostic at this point.  

As for the comments I'm using AWS as just an example since that's what the focus is now.

PR #46, opened 2018-06-19
Improved test db creation

adberglund:
This PR allows for a .env.test file to live in the base of the masu repo with test specific database access credentials. Even if your environement is not currently set to reference your test database it will source that file when creating the test database.  

Example:

```
DATABASE_ENGINE=postgresql
DATABASE_NAME=test
DATABASE_HOST=localhost
DATABASE_PORT=15432
DATABASE_USER=kokuadmin
DATABASE_PASSWORD=''
```

The create_db script now also allows for variable user names to be substituted into the database dump file on load. This means it shouldn't matter which user you created the test database file with, and it will populate the file with your current database user name.


adberglund:
@chambridge I've added a little blurb to the readme under testing to explain .env.test workflow.

PR #49, opened 2018-06-20
Adding Masu Orchestrator

dccurtis:
**Objective**
Masu will periodically doing several tasks.  These include:

1. Getting a current list of accounts setup in Koku
2. Reach out to the cloud providers to get a current cost usage report
3. Process these cost of usage reports to be stored in the database for Koku to present to the UI

The Masu Orchestrator is responsible for executing these tasks.

**Implementation Details**

***Refactoring***
A significant portion of the PR involves moving and refactoring existing code.  Some of the major refactoring changes include:

- Moving the downloader to the `external` module since it involves sourcing data from outside of Masu
- Making a generic ReportDownloader class and interface so that we can easily support multiple providers in the future.

***Asynchronous task preparation***
Report downloads will ultimately be an asynchronous task.  Initially we want to make the processing pipeline to work synchronously.  A separate PR be done to enable celery for downloads.  To prepare for this, a `processor.tasks` module was added to isolate any logic that is intended to be done asynchronously.  The async implementation work will be wrapping the `processor.tasks.download.get_report_files` method in the appropriate celery decorator.

***Items yet to be resolved***
If there are multiple CUR in a given S3 bucket Masu will select the first report it finds.  This is a temporary change until the desired cost usage report can be specified by the user when adding a provider in Koku [Koku Issue #232](https://github.com/project-koku/koku/issues/232)

**Testing**
Unit test is used to verify this change.  Next step is to hook the Orchestrator object into an existing endpoint (status?) or create a new temporary one so that we can invoke a CUR download and verify that the file is downloaded.


dccurtis:
Nope I added external here: https://github.com/project-koku/masu/pull/45

PR #51, opened 2018-06-20
35 report processor woohoo

adberglund:
The report processor... finally.

Have been testing it through with something like 

```
from masu.processor.report_processor import ReportProcessor
file_name = '/Users/aberglund/Downloads/cloudhealth-cur-11.csv.gz'
proc = ReportProcessor('testcustomer', file_name, 1200000, 'gzip')
proc.process()
```


masu/processor/report_processor.py
```diff
@@ -0,0 +1,480 @@
...
+    def __init__(self, schema_name, report_path, cursor_pos=0, compression=''):
+        """Initialize the report processor.
+
+        Args:
+            schema_name (str): The name of the customer schema to process into
+            report_path (str): Where the report file lives in the file system
+            cursor_pos (int): An integer cursor position in the file.
+                The line number to begin processing at.
+            compression (str): How the report file is compressed.
+                Accepted values: '', 'zip', 'gzip'
```
chambridge: You updated the tuple but not the doc string.

masu/processor/report_processor.py
```diff
@@ -0,0 +1,480 @@
...
+            compression (str): The compression format for the file.
+
+        Returns:
+            (file opener, str): The proper file stream handler for the
+                compression and the read mode for the file
+
+        """
+        if compression == '':
+            return open, 'r'
+        elif compression == 'gzip':
```
chambridge: would it make sense to have a constant for the compression types?

masu/processor/report_processor.py
```diff
@@ -0,0 +1,480 @@
...
+
+                    self._update_mappings()
+
+            if self.processed_report.line_items:
+                self._save_to_db()
+                row_count += len(self.processed_report.line_items)
+            self.report_db.close_psycopg2_connection()
+            self.report_db.close_session()
+
+        return self._cursor_pos + row_count
```
dccurtis: So the cursor position is the row in the file?  I was under the impression that this would be the byte offset of the file.  I'll update the docstring in the report stats accessor when we begin integration

adberglund: Yes, my thinking was that since these are CSVs we're never not going to process an entire row. Things get a little whacky if you byte offset a file and then try to wrap the csv package's readers around that offset file.

masu/processor/report_processor.py
```diff
@@ -0,0 +1,480 @@
...
+from os import path
+
+from masu.config import Config
+from masu.database.report_db_accessor import ReportDBAccessor
+from masu.database.reporting_common_db_accessor import ReportingCommonDBAccessor
+from masu.processor.exceptions import MasuProcessingError
+
+LOG = logging.getLogger(__name__)
+
+ALLOWED_COMPRESSIONS = ('', 'gzip')
```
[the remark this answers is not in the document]
adberglund: I meant to bring this up. Because zip as a compression format can contain multiple files, I think it violates our process a single file at a time convention.  We can add it back in here and special case it, or figure out what to do with it elsewhere. Or move all compression handling outside of the processor.

masu/processor/report_processor.py
```diff
@@ -0,0 +1,480 @@
...
+
+
+# pylint: disable=too-few-public-methods
+class ProcessedReport:
+    """Cost usage report transcribed to our database models.
+
+    Effectively a struct for associated database tables.
+    """
+
+    def __init__(self):
```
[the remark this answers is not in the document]
adberglund: Check out https://books.google.com/books?id=Q0s6Vgb98CQC&lpg=PT212&dq=Python+Cookbook+%22Collecting+a+Bunch+of+Named+Items%22&hl=en&pg=PT213#v=onepage&q&f=false

I know it offends your OO sensibilities, but Python really does blur a lot of object oriented lines. Everything is an object, nothing is truly private, duck typing, data encapsulation feels like an afterthought, et cetera. I would argue that getters and setters are not "Pythonic", and in this case they wouldn't be adding anything since I have no checks or modifications to the incoming data. A public attribute that I add to in the processor is all I need. This could be a dictionary of dictionaries and lists that I then just wipe out every time we save to the DB, but I find the attribute syntax on a class more readable.

PR #52, opened 2018-06-21
Provider name being used in place of provider type

dccurtis:
Quick fix to a bug I found while attempting CUR downloads from AWS


masu/external/accounts/db/cur_accounts_db.py
```diff
@@ -49,7 +49,7 @@ def get_accounts_from_source(self):
             auth_credential = provider_accessor.get_authentication()
             billing_source = provider_accessor.get_billing_source()
             customer_name = provider_accessor.get_customer()
-            provider = provider_accessor.get_provider_name()
+            provider = provider_accessor.get_type()
```
chambridge: maybe we should be more explicit with the variable name, `provider_type`?

masu/external/report_downloader.py
```diff
@@ -55,7 +55,7 @@ def _set_downloader(self):
             (Object) : Some object that is a child of CURAccountsInterface
 
         """
-        if self.provider_type == 'aws':
+        if self.provider_type == 'AWS':
```
chambridge: Should we extract this to a constants file?

masu/external/accounts/db/cur_accounts_db.py
dccurtis: Yeah, I agree.  I'll update this

tests/external/accounts/db/test_cur_accounts_db.py
```diff
@@ -38,4 +38,4 @@ def test_get_accounts_from_source(self):
         self.assertEqual(account.get_access_credential(), 'arn:aws:iam::111111111111:role/CostManagement')
         self.assertEqual(account.get_billing_source(), 'test-bucket')
         self.assertEqual(account.get_customer(), 'Test Customer')
-        self.assertEqual(account.get_provider(), 'AWS')
+        self.assertEqual(account.get_provider_type(), 'AWS')
```
chambridge: Could use constant here.

tests/processor/test_orchestrator.py
```diff
@@ -46,7 +46,7 @@ def test_initializer(self):
         self.assertEqual(account.get_access_credential(), 'arn:aws:iam::111111111111:role/CostManagement')
         self.assertEqual(account.get_billing_source(), 'test-bucket')
         self.assertEqual(account.get_customer(), 'Test Customer')
-        self.assertEqual(account.get_provider(), 'AWS')
+        self.assertEqual(account.get_provider_type(), 'AWS')
```
chambridge: Can use constant here.

PR #56, opened 2018-06-22
Integrating ReportProcessor into the orchestration pipeline

dccurtis:
Integrating ReportProcessor into the Masu Orchestrator.  The complete download and process can be triggered synchronously with the /api/v1/download/ end point.

Test Results:
[aws_download_process_test copy.txt](https://github.com/project-koku/masu/files/2128230/aws_download_process_test.copy.txt)


masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -116,7 +114,7 @@ def _get_report_path(self, date_time):
 
         """
         report_date_range = utils.month_date_range(date_time)
-        return '{}/{}/{}'.format(self.bucket_prefix,
+        return '{}/{}/{}'.format(self.report['S3Prefix'],
```
chambridge: it would be safer to use `.get()` in case changes occur to the structure

masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -194,5 +193,8 @@ def download_report(self, date_time):
             stats_recorder.update(etag=etag)
             stats_recorder.commit()
 
-            files.append(file_name)
-        return files
+            report_dictionary['file'] = file_name
+            report_dictionary['compression'] = self.report['Compression']
```
chambridge: it would be safer to use `.get()` in case changes occur to the structure

masu/processor/orchestrator.py
```diff
@@ -69,5 +72,27 @@ def download_curs(self):
                                                      access_credential=credentials,
                                                      report_source=source,
                                                      provider_type=provider)
+            for report_dict in reports:
+                cur_request = CURProcessRequest()
+                cur_request.schema_name = account.get_schema_name()
+                cur_request.report_path = report_dict['file']
```
chambridge: it would be safer to use `.get()` in case changes occur to the structure

tests/external/downloader/aws/test_aws_report_downloader.py
```diff
@@ -288,13 +293,23 @@ def test_download_report(self):
 
         # actual test
         out = self.report_downloader.download_report(fake_report_date)
+        files_list = []
+        for cur_dict in out:
+            files_list.append(cur_dict['file'])
```
chambridge: Maybe have a constant for these?

tests/processor/test_orchestrator.py
```diff
@@ -51,19 +52,42 @@ def test_initializer(self):
...
+        request1 = CURProcessRequest().report_path = '/test/path/file.csv'
+        requests.append(request1)
+        print((request1))
+        request2 = CURProcessRequest().report_path = '/test/path/file2.csv'
+        requests.append(request2)
+        print((request2))
+
+        orchestrator = Orchestrator()
+        orchestrator._processing_requests = requests
+        orchestrator.process_curs()
```
chambridge: seems to execute code and print stuff but not really assert anything.

tests/processor/test_orchestrator.py
```diff
@@ -51,19 +52,42 @@ def test_initializer(self):
...
+
+    @patch('masu.processor.tasks.process.process_report_file', return_value=None)
+    @patch('masu.external.accounts_accessor.AccountsAccessor.get_accounts', return_value=[])
+    def test_process_curs_not_accounts(self, mock_task, mock_accounts_accessor):
+        """Test downloading cost usage reports with no pending requests."""
+        orchestrator = Orchestrator()
+
+        orchestrator.process_curs()
```
chambridge: seems to execute code and print stuff but not really assert anything.

dccurtis: same as above

tests/processor/tasks/test_process.py
```diff
@@ -0,0 +1,43 @@
...
+
+from unittest.mock import patch
+
+from masu.processor.cur_process_request import CURProcessRequest
+from masu.processor.tasks.process import process_report_file
+
+from tests import MasuTestCase
+
+
+class process_report_fileTests(MasuTestCase):
```
adberglund: class name should be `camelCase`'d could be `ProcessTests` to cover all functions in process.py

tests/external/downloader/aws/test_aws_report_downloader.py
dccurtis: I'll clean this up

masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -194,5 +193,8 @@ def download_report(self, date_time):
             stats_recorder.update(etag=etag)
             stats_recorder.commit()
 
-            files.append(file_name)
-        return files
+            report_dictionary['file'] = file_name
+            report_dictionary['compression'] = self.report.get('Compression')
```
adberglund: This is more a fault on my end, because I think my solution was too rigid. Right now if compression is not either `''` or `'gzip` the processor will fail. I think I should have allowed `None` to mean no uncompressed as well. I can turn around with a quick fix for that.

tests/processor/test_orchestrator.py
adberglund: I don't mind deferring until we do make it async, but just a note on testing celery tasks. It's really just testing the synchronous version of the task, so maybe there won't be too much change.

masu/processor/orchestrator.py
```diff
@@ -69,5 +72,27 @@ def download_curs(self):
...
+        Process downloaded cost usage reports.
+
+        Args:
+            None
+
+        Returns:
+            None.
+
+        """
+        for request in self._processing_requests:
```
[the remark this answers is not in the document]
dccurtis: Yeah I know.. Can you think of a better term?  It's still a request for processing.  Need to check the thesaurus :)

masu/external/downloader/aws/aws_report_downloader.py
dccurtis: OK sounds good

masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -74,10 +74,8 @@ def __init__(self, customer_name, auth_credential, cur_source, report_name=None,
         if not report:
             raise MasuConfigurationError('Cost and Usage Report definition not found.')
 
-        report_to_download = report.pop()
-        self.region = report_to_download['S3Region']
+        self.report = report.pop()
```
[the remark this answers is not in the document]
dccurtis: I agree that the logic isn't very clear but I think it's covered with the MasuConfigurationError being thrown right above this.

It's gated by making sure that report_name is in the definitions returned.  report_name can be specified or a default value for the first one we come across.  If there are no reports in the bucket then self.report_name would be None.  Next it would not match any report_defs returned by get_cur_report_definitions.  So report would be None and we would throw the error.

Did I miss something here?

masu/processor/orchestrator.py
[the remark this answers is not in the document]
adberglund: Nah, it's fine, just had the thought as I was looking through

masu/external/downloader/aws/aws_report_downloader.py
adberglund: You are right. For some reason I though uncompressed was an option, but they will always compress the files.

PR #60, opened 2018-06-26
Very quick and dirty async implementation for report processing

adberglund:
## NOT TO BE MERGED

This is really here just as a proof of concept of the simplest way to get what we had async. It's short-cutty and dirty, so don't take this as a implied endorsement of doing things this way on my part. I chained the tasks to short cut out the results back end. The periodic task hasn't been tested at all. Et cetera.

How to run this:

1) Make sure Postgres and RabbitMQ are running in OpenShift

2) Setup a provider in koku pointed to a valid roleARN and cost usage report S3 bucket (I'm using the `cost-usage-bucket` and roleARN in our/Chris's AWS account). 

3) run `make oc-serve` from masu

4) Hit `http://127.0.0.1:5000/api/v1/download/` in a browser

5) Run `celery worker -A masu.celery.worker:CELERY --loglevel=info` in another terminal window

### DIFF
--- a/masu/api/download.py
+++ b/masu/api/download.py
@@ -34,13 +34,17 @@ class DownloadView(View):
     def dispatch_request(self):
         """Packages response for class-based view."""
         orchestrator = Orchestrator()
-        files = orchestrator.prepare_curs()
-        orchestrator.process_curs()
+        orchestrator.prepare_curs()
+        # orchestrator.process_curs()
 
-        downloaded_results_msg = 'Files downloaded and processed: {}'.format(str(files))
-        logger.info(downloaded_results_msg)
+        # downloaded_results_msg = 'Files downloaded and processed: {}'.format(str(files))
+        # logger.info(downloaded_results_msg)
+
+        # response = {
+        #     'files': files,
+        # }
 
         response = {
-            'files': files,
+            'message': 'Queued report processing tasks.'
         }
         return jsonify(response)
--- a/masu/celery/tasks.py
+++ b/masu/celery/tasks.py
@@ -16,96 +16,22 @@
 #
 """Celery task definitions."""
 
-# pylint: disable=unused-argument, fixme, unused-import
-# FIXME: temporary module-wide disable until tasks are fully implemented.
-
+import datetime
+import logging
 from unittest import mock
 
+from celery.task import periodic_task
 from celery.utils.log import get_task_logger
 
-from masu import create_app
-from masu.celery import create_celery
-from masu.processor.exceptions import MasuProcessingError, MasuProviderError
+from masu.processor.tasks.download import get_report_files
+from masu.processor.tasks.process import process_report_file
+from masu.processor.orchestrator import Orchestrator
 
-MASU = create_app()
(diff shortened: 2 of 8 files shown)


masu/celery/tasks.py
```diff
@@ -16,96 +16,22 @@
...
-from masu import create_app
-from masu.celery import create_celery
-from masu.processor.exceptions import MasuProcessingError, MasuProviderError
+from masu.processor.tasks.download import get_report_files
+from masu.processor.tasks.process import process_report_file
+from masu.processor.orchestrator import Orchestrator
 
-MASU = create_app()
-MASU.app_context().push()
-CELERY = create_celery(MASU)
```
adberglund: @blentz If I did not misunderstand, I think this is the circular-y parts that you referenced? If so, we don't need to instantiate an app or a celery instance here, we can define a task, and the `autodiscover_tasks` call in the app factory should handle associating the tasks.

masu/api/download.py
```diff
@@ -34,13 +34,17 @@ class DownloadView(View):
...
-        downloaded_results_msg = 'Files downloaded and processed: {}'.format(str(files))
-        logger.info(downloaded_results_msg)
+        # downloaded_results_msg = 'Files downloaded and processed: {}'.format(str(files))
+        # logger.info(downloaded_results_msg)
+
+        # response = {
+        #     'files': files,
+        # }
 
         response = {
```
dccurtis: This is where I was saying we could block until async results are complete.  Alternatively, we could also use flower to verify completion or make other test endpoints to get the status of pending results.  If we went the test endpoint route we would return some unique identifier that would be the path for status.

```
/api/v1/download => 200 response body: { [pending_async_req1, pending_async_req2,...]}
/api/v1/download/status/{} => i.e. /api/v1/download/status/pending_async_req1 => 200 response body: { 'files': files} or whatever
```

masu/celery/tasks.py
dccurtis: I am also curious where the circular dependency was in the design.  It was one thing I was very aware of and made every effort to avoid that when I was working on the orchestrator and integrating the downloader/processor.  If I missed something please let me know.

masu/api/download.py
blentz: Celery provides a results backend where it reports the results of each task. 

If we need to check the results of the task, we should use the facility provided by Celery.

masu/celery/tasks.py
blentz: @adberglund - That's correct. This file was the source of circular imports.

blentz: @dccurtis The circular dependency ends up being because of how Celery uses decorators to mark tasks.

To get access to the decorators, you need a celery instance.
To get a celery instance, you can import the `CELERY` instance right there.
But, importing CELERY is a problem when, say, the `api.download` view has already imported the `processor.tasks` module that you're trying to mark with the `@CELERY.task` decorators.

blentz:
We've merged #72, this can be closed.

PR #70, opened 2018-06-28
Remove redundant code & reorganize to begin establishing some project conventions.

blentz:
This cleans up several things.

Highlights:
- Stopped tests from using `pass` in `setUp()`
- Re-organized code into a structure that will enable dynamic introspection in the future.
- Created a common space for interfaces to live
- Removed a bunch of redundancies and simplified several code paths
- Moved several string constants into Config module
- Migrated to using application-wide exceptions and moved exceptions to a central location
- Fixed several bugs


masu/config.py
```diff
@@ -66,3 +66,11 @@ class Config(object):
     REPORT_PROCESSING_BATCH_SIZE = 10000
 
     AWS_DATETIME_STR_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+
+    # Provider Types
+    AMAZON_WEB_SERVICES = 'AWS'
```
adberglund: There are constants, and not likely to be configurable values. I don't think they belong in a config file.

masu/processor/downloader.py
```diff
@@ -16,36 +16,34 @@
...
         self.credential = access_credential
-        self.cur_source = report_source
+        self.report_source = report_source
         self.report_name = report_name
         self.provider_type = provider_type
         try:
             self._downloader = self._set_downloader()
         except Exception as err:
-            raise ReportDownloaderError(str(err))
+            raise MasuProviderError(str(err))
```
adberglund: Why did we move to a less specific error in this case?

masu/processor/__init__.py
```diff
@@ -15,10 +15,3 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 """Masu Processor."""
-import logging
-
-from masu.external import GZIP_COMPRESSED, UNCOMPRESSED
-
-LOG = logging.getLogger(__name__)
-
-ALLOWED_COMPRESSIONS = (UNCOMPRESSED, GZIP_COMPRESSED)
```
adberglund: These were fine as constants. Don't remove.

masu/processor/downloader.py
```diff
@@ -16,36 +16,34 @@
...
+from masu.providers.aws.downloader import AWSReportDownloader
 
 
 # pylint: disable=too-few-public-methods
 # pylint: disable=too-many-arguments
 class ReportDownloader:
-    """Interface for masu to use to get CUR accounts."""
+    """Top-level interface for masu to use for downloading reports."""
+
+    # FIXME: this class interface will change once we teach it how to properly
```
adberglund: FIXME implies that this is broken as is. Is it broken?

dccurtis:
Thanks for sharing this so that the entire team can weigh in.  I'll summarize our conversation for the benefit of others with some new thoughts:

- I'm still going through the change and will provide feedback when I'm ready (it's a big change so it's going to take some time.). To be perfectly honest, it's not my number 1 priority at this point in time since I have not been told of anything thats specifically broken or blocking.
- I am 100% on board with having a general discussion on the goals of the design and conventions we want to follow.  I am working on drafting my thoughts of how the lay of the land is today and would prefer that we get aligned on what engineering priorities and patterns we want to follow to meet the project's short term and long term goals before combing through a massive PR (not to mention nothing is actually broken)
- FYI: In general, I won't spend a lot of time reviewing PR code until all tests are passing.
- As a result of the team alignment on what's important and what's not important to make our project successful I think we should piecemeal the changes in.  It will make sure the refactoring is done in a systematic and testable way that will benefit not only our understanding of how the code works but also will be much easier to catch regressions.
- If *anything* is blocking you with making progress with the immediate work for this sprint please bring it up so we can address things

dccurtis:
To clarify my comment about not spending a lot of time reviewing code with failing tests.  I will still be reviewing the code :). I think it will be very helpful to get your perspective on things.  It's just going to be more general statements then drilling too deep on specifics

masu/providers/database/accessors.py
```diff
@@ -14,11 +14,16 @@
...
-"""Initializer for external module."""
+"""Standardized namespacing entry-point for DB accessors."""
 
-# Provider Types
-AMAZON_WEB_SERVICES = 'AWS'
-
-# Compression formats
-UNCOMPRESSED = ''
-GZIP_COMPRESSED = 'GZIP'
+# pylint: disable=unused-import
```
adberglund: If we have different accessors they should just be in an accessors module not imported into an accessors file like this. Then you import with the standard package and module import process. See https://docs.python.org/3/tutorial/modules.html#packages

blentz:
I don't agree with the "nothing is actually broken" statement. While I was working through this, I found a few different bugs that are also getting fixed by this change. One reason for some of these bugs to go uncaught is that there are a couple of tests that are missing assertions.

The main blocker for me is that there were several redundancies introduced into the codebase as things were getting moved around. So, before I move forward with introducing async, I want to get that cleaned up before it becomes a larger issue.

blentz:
Doh. This PR is from the wrong branch. Closing this out and recreating.

dccurtis:
@blentz Lets go ahead and address the bugs you found in separate well-scoped PRs
I should have been more clear on that statement.  I didn't mean it was "bug-free" I meant the overall design was working on a macro-level.  Software at this early stage is going to have behavior that is not completely nailed down yet.  For me the orchestrator was that thing that was very undefined and I kept it down to the bare minimum to download and process files

PR #71, opened 2018-06-28
Remove redundant code & reorganize to begin establishing some project conventions.

blentz:
This cleans up several things.

Highlights:

- Stopped tests from using `pass` in `setUp()`
- Re-organized code into a structure that will enable dynamic introspection in the future.
- Created a common space for interfaces to live
- Removed a bunch of redundancies and simplified several code paths
- Moved several string constants into Config module
- Migrated to using application-wide exceptions and moved exceptions to a central location
- Fixed several bugs


dccurtis:
For those of you that missed the other PR, I've got comments that are still relevant here
https://github.com/project-koku/masu/pull/70

Last comment where 70 left off:
@blentz Lets go ahead and address the bugs you found in separate well-scoped PRs
I should have been more clear on that statement. I didn't mean it was "bug-free" I meant the overall design was working on a macro-level. Software at this early stage is going to have behavior that is not completely nailed down yet. For me the orchestrator was that thing that was very undefined and I kept it down to the bare minimum to download and process files

blentz:
I disagree. The macro-level design also had significant issues.

- Creating classes where a dict was sufficient
- Prematurely defining interfaces when we haven't even defined the multi-provider use case. 
- Adding redundant code paths so that half the code called one function and the new code called a new function, which was just a no-op wrapper around the old function.

masu/processor/tasks/process.py
```diff
@@ -39,18 +39,18 @@ def process_report_file(process_request):
             ' schema_name: {},'
             ' report_path: {},'
             ' compression: {}')
-    log_statement = stmt.format(process_request.schema_name,
-                                process_request.report_path,
-                                process_request.compression)
+    log_statement = stmt.format(process_request['schema_name'],
```
chambridge: `.get()` would be safer.

blentz:
After some offline discussion, we've made a decision to break this PR up into smaller pieces to make it easier to review. This will also let us prioritize some pieces ahead of others that need additional discussion.

I'll leave this PR open for now. I'll close it as soon as I start creating the new PRs.

dccurtis:
OK I went through the entire review and was focusing on the following items:
1) What capabilities have been added or removed
2) How does it deviate from the existing fundamental flow.  Especially looking out for what could have been the bugs you said you found that was hindering the async work that was resolved here.
3) General organization changes

**One sentence summary:**
Flow and general approach is exactly the same, but we would lose the ability easily source account information from anything except the database.

On the whole, I’m fine with it. Really sad about removing capabilities that were not hindering a thing however…

**_Question_**: Can you point out where you ran into problems implementing the async work and how this PR addresses it?

**Summary of my notes:**
Random thoughts while reading code:
* name changes: **OK**
* module/directory movement.  **Overall fine, see notes below.**
* exceptions being moved from classes to an overall project level.  Need to see how this will play out, but **OK**
* CURProcessRequest is now a dictionary. **OK**
* Concept of ‘external’ is gone and has instead collapsed into a provider folder.  Some details:
external/downloader/aws/aws_report_downloader.py -> providers/aws/downloader.py (this is aws specific still even though it’s called downloader.py, but it’s in the directory for aws so, **OK**.  DownloadInterface abc name changed, but we can still easily plug in other downloaders, yay!  _Did you miss that one?_
* external/report_downloader.py -> processor/downloader.py.  In my mind, the processor and downloader serve two very different tasks and shouldn’t be grouped.  **OK, I guess**
* external/accounts: completely gone and instead changed to a database only way to source account information (CURAccountsDB became Account()).
* CURAccountsDB() (external/accounts/db/cur_accounts_db.py) => Account() (processor/account.py).  To me sourcing accounts access details has nothing to do with processing cost usage reports.  I don’t understand why it’s in the processor module.
* Account() fills in a list of dictionaries instead of CostUsageReportAccount object.  **OK no problem.**
* Removed all associated account sourcing abstraction (CostUsageReportAccount).  Lost capability of sourcing accounts from anything but the database (i.e. no easy REST API plugin anymore).
* Database accessors were moved to providers (masu/database -> masu/providers/database) with some name changes.  These database accessors are accessing the Koku database which generally has nothing to do with the provider itself (sure there is a provider table in that database, but the accessor is for the database itself, thats why I had it in a top level directory, to isolate db access calls to be used across the project).  **Don’t agree with it but I can live with it.**

PR #72, opened 2018-07-05
implement async tasks with celery

blentz:


chambridge:
Restarted timed out builds in case it was a Travis issue.

masu/celery/__init__.py
```diff
@@ -23,11 +23,17 @@ def create_celery(app):
     """Create Celery app object using the Flask app's settings."""
     celery = Celery(
         app.import_name,
-        backend=app.config.get('CELERY_RESULT_BACKEND'),
+        # backend=app.config.get('CELERY_RESULT_BACKEND'),
```
chambridge: Any reason to leave this commented out in the code vs. deleted? Or should it not commented out?

masu/config.py
```diff
@@ -61,8 +62,14 @@ class Config(object):
 
     # Celery settings
     CELERY_BROKER_URL = f'amqp://{RABBITMQ_HOST}:{RABBITMQ_PORT}'
-    CELERY_RESULT_BACKEND = f'db+postgresql://{DB_USER}:{DB_PASSWORD}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
+    # CELERY_RESULT_BACKEND = f'amqp://{RABBITMQ_HOST}:{RABBITMQ_PORT}'
```
chambridge: Not sure why we have duplicate commented out environment values here.

masu/config.py
```diff
@@ -61,8 +62,14 @@ class Config(object):
...
+    # pylint: disable=line-too-long
+    # CELERY_RESULT_BACKEND = f'db+postgresql://{DB_USER}:{DB_PASSWORD}@{DB_HOST}:{DB_PORT}/{DB_NAME}'
 
     REPORT_PROCESSING_BATCH_SIZE = 10000
 
     AWS_DATETIME_STR_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+
+    # The interval for scanning for new reports.
+    # This needs to be a timedelta object.
+    REPORT_CHECK_INTERVAL = datetime.timedelta(minutes=5)
```
chambridge: Seems pretty often, thought we were going for hourly based on our last discussion. Not a blocker just was my expectation.

I also think we should drive this with an environment variable that defaults to a value if not present.

blentz: This is the "development value" so that I'm not waiting around forever to see if the scheduled task runs. 

My plan was to reset it to something sane once I actually got the scheduled tasks running. They don't currently work. 

Once the basic async stuff is merged, I'll focus on getting scheduled tasks also working.

masu/celery/__init__.py
blentz: I suspect at some point, we'll actually want to use the result backend to check task status. So, for now, I'm leaving it there just so whoever comes along to implement that doesn't have to re-wire that stuff up.

Issue #67, opened 2018-06-26
Lookup provider from SNS request information


chambridge:
When an SNS update comes in we will get the S3 bucket name and cost usage definition key. We need to be able to look up the provider from the S3 bucket name so we can trigger the download and processing appropriately.

If the provider is not found within the in memory data, assume the data is stale and call to obtain the updated data. If the provider is still not found log the info as an error.

______
Associated with: https://github.com/project-koku/masu/issues/7

dccurtis:
The situation of having stale provider data will not occur with the current implementation.  We are creating a new orchestrator each time we download/process.  During orchestrator initialization a new list of accounts are created each time.

This may be a problem if we ever move to a model where the orchestrator lives forever and handles requests as they come in.  I'm closing this for now since there is no issue with the current implementation.

PR #73, opened 2018-07-05
Handle SNS notifications. Closes #66, #67, #7

### LINKED ISSUE #66: Create endpoint to handle an SNS request
Add an API endpoint that will handle an SNS update of a customers cost and usage report data.

Its expected that the message will contain the bucket information and they key to report definition (i.e. manifest file)

______
Associated with: https://github.com/project-koku/masu/issues/7

dccurtis:
Closes #66, #67, and #7.

**Objective**
Amazon Simple Notification System can be used to alert Masu whenever a S3 bucket that it is monitoring has new data to process.  The alert comes in the form of an HTTP POST to a pre-deifined masu endpoint.

The HTTP endpoint will be subscribed to an AWS SNS Topic and whenever a S3 bucket pushes a message to the Topic (via. S3 Events configuration) then AWS will issue a POST request to the subscribed notification endpoint.

The purpose of this is to ensure that masu has downloads and processes cost usage reports in a timely manner.

SNS is optional for the end-user, if it is not configured then masu will rely on the regular polling mechanism to ensure reports are downloaded and processed.

**Implementation Details**
* AWS Topic Subscription: While there are several ways to subscribe to a topic, for this change I decided to allow for HTTP end point subscription to be driven from the AWS Dashboard.  This will be a manual, one-time setup for the service administrator.  Masu should be running when the endpoint is registered to the Topic.  Masu will automatically confirm the subscription
* The Topic can be configured to send notifications for ```ObjectCreate (All)``` to cover the most broad notification use case.  Masu will filter out all non-top-level Manifest files (top-level manifest file changes each time any file changes).
* The event handling logic works like this: 1) Receive notification 2) Extract S3 bucket name (globally unique) 3) Pass bucket name to Masu Orchestrator 4) Orchestrator finds necessary S3 access credentials 5) Does the normal download/process workflow

**Areas of improvement**
* We should use an environment variable or something similar to specify what Topic we are expecting to see when confirming a HTTP subscription request.  As it stands with this PR, Masu will accept *any* Topic endpoint subscription.
* Still need to verify the authenticity of the AWS messages by implementing the signature check.

**Testing**
I've included the logs and basic setup work flow followed to verify the SNS handling functionality.  This will be the basis for a document I'm putting together to explain the setup.

[steps.txt](https://github.com/project-koku/masu/files/2167711/steps.txt)


dccurtis:
OK, I tried to resolve the conflicts on GitHub from @blentz's latest merge, proved unsuccessful.  Stand by for updates to make this work with the latest async changes.

masu/external/notifications/aws/aws_notification_handler.py
```diff
@@ -0,0 +1,174 @@
...
+                                           an HTTP endpoint
+        Raises:
+            (AWSNotificationHandlerError): Is raised when the notification contains a
+                                           message type other than 'SubscriptionConfirmation'
+                                           or 'Notification'
+
+        """
+        msg_type = None
+        for header in self._headers:
+            if header[0] == AWS_SNS_HEADER_MSG_TYPE:
```
blentz: This could be simplified down to 
`filter(lambda header: header[0] == AWS_SNS_HEADER_MSG_TYPE, self._headers)`

masu/external/notifications/aws/aws_notification_handler.py
```diff
@@ -0,0 +1,174 @@
...
+                if re.findall(self.UUID_REGEX, object_key):
+                    msg = 'Ignoring non-toplevel manifest file: {}'.format(object_key)
+                    raise NotificationInterfaceFilter(msg)
+
+            if not object_key.endswith('Manifest.json'):
+                msg = 'Ignoring non-manifest file: {}'.format(object_key)
+                raise NotificationInterfaceFilter(msg)
+
+            # There must be a more reliable way of getting bucket name..
+            s3_bucket_name = message_dict['Records'][0]['s3']['bucket']['name']
```
blentz: Just to respond to the comment about digging into deeply nested data...

If we need to do this often enough, it might make sense to use something like `objectpath`. See: http://objectpath.org/

masu/external/notifications/notification_interface.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Notification handler interface to be used by Masu."""
+
+from abc import ABC, abstractmethod
+
+
+class NotificationInterfaceFilter(Exception):
```
blentz: This should be in `exceptions.py`

masu/processor/orchestrator.py
```diff
@@ -35,16 +35,40 @@ class Orchestrator():
 
     """
 
-    def __init__(self):
+    def __init__(self, billing_source=None):
```
blentz: This seems counter to our discussions that orchestrator is expected to enumerate across all accounts and all billing sources.

What's the intent here?

masu/processor/orchestrator.py
```diff
@@ -35,16 +35,40 @@ class Orchestrator():
...
         """
         Orchestrator for processing.
 
         Args:
-            None
+            billing_source (String): Individual account to retrieve.
         """
-        self._accounts = AccountsAccessor().get_accounts()
+        self._accounts = self.get_accounts(billing_source)
```
blentz: Isn't this a static method?  If so, it should be called like this: `Orchestrator.get_accounts()`.

This looks like you're calling an instance method.

masu/processor/orchestrator.py
```diff
@@ -35,16 +35,40 @@ class Orchestrator():
...
         self._processing_requests = []
 
+    @staticmethod
+    def get_accounts(billing_source=None):
```
blentz: It's not clear to me why this belongs in the Orchestrator. Can you help me understand?

dccurtis:
Next steps are to extract the AWS Region from the subscription confirmation request body and supply it to the SNS client.  Then I'm going to do a sanity test to make sure everything still works with the latest async change.

tests/api/test_notification.py
```diff
@@ -0,0 +1,162 @@
...
+                     "Timestamp" : "2018-07-04T18:04:41.844Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "J7rG3Jd9SR0So5kbkOHgfSJvOVckW5paIEc7MTTraVe1DxKzImpwHAabEAqj1dI9daPe8SlIEK92HocNjgIj9Ox3n/KE+aN871QSQ5ISG1P54uqnkaw+fwfjsFwVGnUsNQ50ntaM9JlsK6VarwnBKdEjP/EAIS00R+bueJQ0Yws8Xhvyzmw0zASRminho5EpMNpucUAY4oKjgn3Bir7MU4d/d31ZZedrdKb35XcKpIs1lZj/MPHOWRM2NDG+6AhSlEbZ94IIi9ycqoRgDn16hIBVLMbFRSdbeRjQqYUxru3Inp5upRqOUTw0yNPLPbdUD/+qiUYs9A1Mc+VPZ/M0PA==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageReportUpdateTopic:7d549a1b-502f-47f7-9469-9ed18671e76a"}
+        body = json.dumps(body_dict)
```
blentz: It would be good to add some fuzziness to this test data. e.g. use Faker to generate the message timestamp and uuids, randomly pick regions, etc.

Not critical, but definitely nice to help improve the test quality.

tests/api/test_notification.py
```diff
@@ -0,0 +1,162 @@
...
+        response = self.client.post('/api/v1/notification/', headers=header, data=body)
+        body = response.json
+
+        self.assertEqual(response.status_code, 204)
+        self.assertEqual(response.headers['Content-Type'], 'text/html; charset=utf-8')
```
blentz: We should probably also assert that the message body or message uuid is what we expect it to be.

tests/api/test_notification.py
```diff
@@ -0,0 +1,162 @@
...
+                     "Timestamp" : "2018-07-04T18:04:41.844Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "J7rG3Jd9SR0So5kbkOHgfSJvOVckW5paIEc7MTTraVe1DxKzImpwHAabEAqj1dI9daPe8SlIEK92HocNjgIj9Ox3n/KE+aN871QSQ5ISG1P54uqnkaw+fwfjsFwVGnUsNQ50ntaM9JlsK6VarwnBKdEjP/EAIS00R+bueJQ0Yws8Xhvyzmw0zASRminho5EpMNpucUAY4oKjgn3Bir7MU4d/d31ZZedrdKb35XcKpIs1lZj/MPHOWRM2NDG+6AhSlEbZ94IIi9ycqoRgDn16hIBVLMbFRSdbeRjQqYUxru3Inp5upRqOUTw0yNPLPbdUD/+qiUYs9A1Mc+VPZ/M0PA==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageReportUpdateTopic:7d549a1b-502f-47f7-9469-9ed18671e76a"}
+        body = json.dumps(body_dict)
+        response = self.client.post('/api/v1/notification/', headers=header, data=body)
+        body = response.json
+
+        self.assertEqual(response.status_code, 204)
```
blentz: If we're submitting an erroneous message, the API shouldn't return a 200-code. It should be a 400-code.

tests/api/test_notification.py
```diff
@@ -0,0 +1,162 @@
...
+                     "Timestamp" : "2018-07-04T18:04:41.844Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "J7rG3Jd9SR0So5kbkOHgfSJvOVckW5paIEc7MTTraVe1DxKzImpwHAabEAqj1dI9daPe8SlIEK92HocNjgIj9Ox3n/KE+aN871QSQ5ISG1P54uqnkaw+fwfjsFwVGnUsNQ50ntaM9JlsK6VarwnBKdEjP/EAIS00R+bueJQ0Yws8Xhvyzmw0zASRminho5EpMNpucUAY4oKjgn3Bir7MU4d/d31ZZedrdKb35XcKpIs1lZj/MPHOWRM2NDG+6AhSlEbZ94IIi9ycqoRgDn16hIBVLMbFRSdbeRjQqYUxru3Inp5upRqOUTw0yNPLPbdUD/+qiUYs9A1Mc+VPZ/M0PA==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageReportUpdateTopic:7d549a1b-502f-47f7-9469-9ed18671e76a"}
+        body = json.dumps(body_dict)
```
blentz: I'm seeing the same (or is it just similar?) test data copied and pasted multiple times. Any way we can collect the common values and just define them once to DRY this up?

tests/external/downloader/aws/test_aws_report_downloader.py
```diff
@@ -262,6 +263,27 @@ def test_download_file(self):
...
+            _, _ = self.report_downloader.download_file(missing_key)
+        expected_err = 'Unable to find {} in S3 Bucket: {}'.format(missing_key, self.fake_bucket_name)
+        self.assertEqual(expected_err, str(error.exception))
+
+    @mock_s3
+    def test_download_file_other_error(self):
+        fake_object = self.fake.word().lower()
+        # No S3 bucket created
+        with self.assertRaises(AWSReportDownloaderError) as error:
+            _, _ = self.report_downloader.download_file(fake_object)
```
blentz: Don't assign variables if you aren't going to use them. Also, don't use `_` just to avoid pylint complaining about unused-variables.

Let's simplify this down to:

```
        with self.assertRaises(AWSReportDownloaderError) as error:
            self.report_downloader.download_file(fake_object)
```

tests/external/downloader/aws/test_aws_report_downloader.py
```diff
@@ -262,6 +263,27 @@ def test_download_file(self):
...
+        self.assertTrue('NoSuchBucket' in str(error.exception))
```
blentz: Is `error.exception` an Iterable?  If it's not, this should probably be an `assertEqual`

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "Timestamp" : "2018-07-05T12:29:27.641Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "L8VmRMT55sv326Rt1HBawA8hPKR0plaBfvI8hlZNm2nhaMkVzvFkvKiMZucfQ7bYWaYqVF+xzvqkaRnMw8/uC4QcrM5DK5HcXYfAJCeI8t32jTg3Kcdn3SGBiyxcBWYt8qSXL7403ho4vIVpCHRuX6K2ZydS7sxx6umr/nanaOUGnkDeC91GFoUMslNw7yITc0YunPwg8GcvKgLhsQr2d7C5I3zDprA+Ds29voLN6v66hH0uR0LEfDq9Ky0Dgii4UebYUgKBCWSy5ETsRHqug5tPNzv9afexjdXrjEUYbvYZO+RRDPztwGjyHw6CL35B85FLZ7UDPLeRysZq47gb+Q==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem"}
+        body = json.dumps(body_dict)
+        try:
+            _ = AWSNotificationHandler(headers, body, validation=False)
```
blentz: As above: unused-variable

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "SubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=ConfirmSubscription&TopicArn=arn:aws:sns:us-east-1:589179999999:CostUsageNotificationDemo&Token=2336412f37fb687f5d51e6e241da92fd72d769c811cb257f2ca8bbb9e329dd2eb427b0935ce7fe84f91e14e5994d2d9039b3aade5b4f16dd5ab3979b05751962ceed2c7e193e7b7797630c31eec29477949064522feb104a311b44e1268ba7bdfdc06d7e6f39174dbb69326a379d06cbbc96adcdc24939e36892e024aa0e96d2",
+                     "Timestamp" : "2018-07-05T12:29:27.641Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "L8VmRMT55sv326Rt1HBawA8hPKR0plaBfvI8hlZNm2nhaMkVzvFkvKiMZucfQ7bYWaYqVF+xzvqkaRnMw8/uC4QcrM5DK5HcXYfAJCeI8t32jTg3Kcdn3SGBiyxcBWYt8qSXL7403ho4vIVpCHRuX6K2ZydS7sxx6umr/nanaOUGnkDeC91GFoUMslNw7yITc0YunPwg8GcvKgLhsQr2d7C5I3zDprA+Ds29voLN6v66hH0uR0LEfDq9Ky0Dgii4UebYUgKBCWSy5ETsRHqug5tPNzv9afexjdXrjEUYbvYZO+RRDPztwGjyHw6CL35B85FLZ7UDPLeRysZq47gb+Q==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem"}
+        body = json.dumps(body_dict)
```
blentz: As above, DRY this up, if we can.

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            _ = AWSNotificationHandler(headers, body, validation=False)
+        except Exception:
+            self.fail('Unexpected exception')
```
blentz: Why is this try/except block here? Won't the test fail if an exception is thrown anyway?

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "SubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=ConfirmSubscription&TopicArn=arn:aws:sns:us-east-1:589179999999:CostUsageNotificationDemo&Token=2336412f37fb687f5d51e6e241da92fd72d769c811cb257f2ca8bbb9e329dd2eb427b0935ce7fe84f91e14e5994d2d9039b3aade5b4f16dd5ab3979b05751962ceed2c7e193e7b7797630c31eec29477949064522feb104a311b44e1268ba7bdfdc06d7e6f39174dbb69326a379d06cbbc96adcdc24939e36892e024aa0e96d2",
+                     "Timestamp" : "2018-07-05T12:29:27.641Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "L8VmRMT55sv326Rt1HBawA8hPKR0plaBfvI8hlZNm2nhaMkVzvFkvKiMZucfQ7bYWaYqVF+xzvqkaRnMw8/uC4QcrM5DK5HcXYfAJCeI8t32jTg3Kcdn3SGBiyxcBWYt8qSXL7403ho4vIVpCHRuX6K2ZydS7sxx6umr/nanaOUGnkDeC91GFoUMslNw7yITc0YunPwg8GcvKgLhsQr2d7C5I3zDprA+Ds29voLN6v66hH0uR0LEfDq9Ky0Dgii4UebYUgKBCWSy5ETsRHqug5tPNzv9afexjdXrjEUYbvYZO+RRDPztwGjyHw6CL35B85FLZ7UDPLeRysZq47gb+Q==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem"}
+        body = json.dumps(body_dict)
```
blentz: As above, DRY this if we can.

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+        with self.assertRaises(AWSNotificationHandlerError):
+            _ = AWSNotificationHandler(headers, body, validation=False)
```
blentz: As above: unused-variable

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "Timestamp" : "2018-07-03T13:07:40.924Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "VOHSWykhwa3UJ9o0QpZHQm6VyxYXF2SMypzQW4DT7GveCrByef02+Xb/j+Fjh2D7kAyFgwq30l2HJc1W048GJ26Xju+x5YRt2KciU88Rra/L97HZmT3lUO3CBWv72nc157QgE1pazuHymjvB4ARaw8Ga1QW4brYBKeyjSL3q8RMXpELnapCc0OzWZ7/tn0sYNbEnro7ZVOAo2NdUgBrNedV/Cu6zjJ2o3k+bYvH9sHN9qI8hGteSg2nshBugTg/7873QbeN5LMaUljTEDnIBaIKdxRLdlj3Q2N0KlFHyMsu4TumqBY2xzEoCa5k5hx7J5F1nerZb0BikwGNGOHkDvA==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageNotificationDemo:3affeac1-5249-49a0-876d-a163fb53a004"}
+        body = json.dumps(body_dict)
+        handler = AWSNotificationHandler(headers, body)
+        with self.assertRaises(NotificationInterfaceFilter):
+            _ = handler.get_billing_source()
+
```
blentz: DRY and unused-variable here, too.

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "Timestamp" : "2018-07-04T18:04:41.820Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "mDniwbH0VTvDtm4VaZf66Ox3joxBv92lJb4KXEqq1AeVKTfpyCuB5XROU7iw4v2AD97DPr/C9tHIOuuMaDANJxVLor7TfJB7xCA9PYg3963NqCE1uvl5fv7NYhCpl4bhGwKwKBIYvOvWbVxcAQilRWZHDmLAqEmWfzCIPtACWc9xAt0rbL2Zk0TqAhU87QU3dAX63bV6HdNOHMCp1xS5bNVWWWJ09vsUNwSVS4zQ/bAxymffoAMyIEBKlCU8le4khjGlmSIX6S6vsl0rVd9e8hInSmAHN4mcBE7SUz/TrlmwwejP6ysEdHRnHfU41/axMxBZyfH654A2Kne6Vs01mg==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageReportUpdateTopic:7d549a1b-502f-47f7-9469-9ed18671e76a"}
+        body = json.dumps(body_dict)
```
blentz: DRY, please.

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "Timestamp" : "2018-07-04T18:04:40.036Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "gMXR9TvaiZjpuslyjnfj9PdOyOg78qqrz4pyt7Ct7i+EWkNbo1ePKoszJv2Wx0k7joX+irfLiCiROt8lr3TuVEA9ht/lG1hQAUW6T+BAikB6IF5c9uR1PtiP+Kn+qG9FyDbtSdr3Qn0i3PwI9vOIrqXNf5aEm1mJKqdwdWvkQ+pYHr1RYn4a5T6rrUuRicfzhD6qlbq8jfgzhHPLUcn4wQZ9Sdq7buuY8I3jCo/p+CGhkg70HaF64dyuzSZQU7DEaumBuMvasTI50w3PgymF4CM3wBGrzqEc+FFoEAgUjJTO2ri//KfBzhMhUkxhkIzzh5rGgrHfVGpV7gyFJ9vGWw==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageReportUpdateTopic:7d549a1b-502f-47f7-9469-9ed18671e76a"}
+        body = json.dumps(body_dict)
+        handler = AWSNotificationHandler(headers, body)
+        with self.assertRaises(NotificationInterfaceFilter):
+            _ = handler.get_billing_source()
```
blentz: DRY and unused-variable here, too.

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def test_get_billing_source_empty_body(self):
+        headers = [('X-Amz-Sns-Message-Type', 'Notification')]
+        body = u''
```
blentz: IIRC, all strings in Python 3 are unicode. I don't think it matters if you use `''` versus `u''`

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+        handler = AWSNotificationHandler(headers, body)
+        s3_bucket_name = handler.get_billing_source()
+        self.assertEqual(s3_bucket_name, None)
+
+    def test_get_billing_source_wrong_message_type(self):
+        headers = [('X-Amz-Sns-Message-Type', 'RandomType')]
+        body = u''
+        with self.assertRaises(AWSNotificationHandlerError):
+            _ = AWSNotificationHandler(headers, body)
```
blentz: unused-variable

tests/external/notifications/aws/test_aws_notification_handler.py
```diff
@@ -0,0 +1,208 @@
...
+                     "Timestamp" : "2018-07-05T12:29:27.641Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "L8VmRMT55sv326Rt1HBawA8hPKR0plaBfvI8hlZNm2nhaMkVzvFkvKiMZucfQ7bYWaYqVF+xzvqkaRnMw8/uC4QcrM5DK5HcXYfAJCeI8t32jTg3Kcdn3SGBiyxcBWYt8qSXL7403ho4vIVpCHRuX6K2ZydS7sxx6umr/nanaOUGnkDeC91GFoUMslNw7yITc0YunPwg8GcvKgLhsQr2d7C5I3zDprA+Ds29voLN6v66hH0uR0LEfDq9Ky0Dgii4UebYUgKBCWSy5ETsRHqug5tPNzv9afexjdXrjEUYbvYZO+RRDPztwGjyHw6CL35B85FLZ7UDPLeRysZq47gb+Q==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem"}
+        body = json.dumps(body_dict)
+        with self.assertRaises(AWSNotificationHandlerError):
+            _ = AWSNotificationHandler(headers, body, validation=False)
```
blentz: DRY, if possible and unused-variable.

tests/external/test_notification_handler.py
```diff
@@ -0,0 +1,101 @@
...
+                     "Timestamp" : "2018-07-03T13:07:40.924Z",
+                     "SignatureVersion" : "1",
+                     "Signature" : "VOHSWykhwa3UJ9o0QpZHQm6VyxYXF2SMypzQW4DT7GveCrByef02+Xb/j+Fjh2D7kAyFgwq30l2HJc1W048GJ26Xju+x5YRt2KciU88Rra/L97HZmT3lUO3CBWv72nc157QgE1pazuHymjvB4ARaw8Ga1QW4brYBKeyjSL3q8RMXpELnapCc0OzWZ7/tn0sYNbEnro7ZVOAo2NdUgBrNedV/Cu6zjJ2o3k+bYvH9sHN9qI8hGteSg2nshBugTg/7873QbeN5LMaUljTEDnIBaIKdxRLdlj3Q2N0KlFHyMsu4TumqBY2xzEoCa5k5hx7J5F1nerZb0BikwGNGOHkDvA==",
+                     "SigningCertURL" : "https://sns.us-east-1.amazonaws.com/SimpleNotificationService-eaea6120e66ea12e88dcd8bcbddca752.pem",
+                     "UnsubscribeURL" : "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:589179999999:CostUsageNotificationDemo:3affeac1-5249-49a0-876d-a163fb53a004"}
+        body = json.dumps(body_dict)
```
blentz: DRY, please.

tests/external/test_notification_handler.py
```diff
@@ -0,0 +1,101 @@
...
+
+        with self.assertRaises(NotificationHandlerError) as error:
+            _ = NotificationHandler(headers, body)
```
blentz: DRY and unused-variable.

tests/external/test_notification_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        headers = [('X-Wizz-Bang-Message-Type', 'Notification'),
+                   ('Content-Length', 761),
+                   ('Content-Type', 'text/plain; charset=UTF-8'),
+                   ('Connection', 'Keep-Alive'),
+                   ('User-Agent', 'Amazon Simple Notification Service Agent')]
+
+        body_dict = {"Type" : "Notification",
+                     "MessageId" : "869ace9b-691a-5148-a22c-9d2e1e46e3de",
+                     "Subject" : "Wizz Bang Notification",
+                     "Timestamp" : "2018-07-03T13:07:40.924Z"}
```
blentz: Use Faker to add some fuzz, please.

tests/external/test_notification_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        body = json.dumps(body_dict)
+
+        with self.assertRaises(NotificationHandlerError) as error:
+            _ = NotificationHandler(headers, body)
```
blentz: unused-variable

tests/external/test_notification_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        self.assertTrue('Unsupported cloud provider.' in str(error.exception))
```
blentz: Same as before - if error.exception isn't an Iterable, this should probably be an `assertEqual`

tests/external/test_report_downloader.py
```diff
@@ -31,8 +31,8 @@ class FakeDownloader():
 class ReportDownloaderTest(MasuTestCase):
     """Test Cases for the ReportDownloader object."""
 
-    def setUp(self):
-        pass
+    file_list = ['/var/tmp/masu/region/aws/catch-clearly.csv',
+                 '/var/tmp/masu/base/aws/professor-hour-industry-television.csv']
```
blentz: Use Faker, please.

tests/external/test_report_downloader.py
```diff
@@ -63,3 +63,27 @@ def test_invalid_provider_type(self):
                                  report_source='hereiam',
                                  report_name='bestreport',
                                  provider_type='unknown')
+
+    @patch('masu.external.downloader.aws.aws_report_downloader.AWSReportDownloader.__init__', return_value=None)
+    def test_get_current_report(self, fake_downloader):
+        """Test get_current_report function."""
+        downloader = ReportDownloader(customer_name='customer name',
+                                      access_credential='mycred',
```
blentz: I've added a new facility to generate a fake ARN. Please use that here.

I had to squash a number of bugs in the tests because we weren't passing ARNs into the credentials.

tests/external/test_report_downloader.py
```diff
@@ -63,3 +63,27 @@ def test_invalid_provider_type(self):
...
+                                      report_source='hereiam',
+                                      report_name='bestreport',
+                                      provider_type=AMAZON_WEB_SERVICES)
+        with patch.object(AWSReportDownloader, 'download_current_report', return_value=self.file_list):
+            files = downloader.get_current_report()
+            self.assertEqual(len(files), len(self.file_list))
```
blentz: The lengths of the lists may be the same, but are the contents of those lists also the same?

tests/external/test_report_downloader.py
```diff
@@ -63,3 +63,27 @@ def test_invalid_provider_type(self):
...
+    def test_get_current_report_error(self, fake_downloader):
+        """Test get_current_report function with error."""
+        downloader = ReportDownloader(customer_name='customer name',
+                                      access_credential='mycred',
+                                      report_source='hereiam',
+                                      report_name='bestreport',
+                                      provider_type=AMAZON_WEB_SERVICES)
+        with patch.object(AWSReportDownloader, 'download_current_report', side_effect=Exception('some error')):
+            with self.assertRaises(ReportDownloaderError):
+                _ = downloader.get_current_report()
```
blentz: unused-variable

tests/external/test_report_downloader.py
```diff
@@ -63,3 +63,27 @@ def test_invalid_provider_type(self):
...
+                                      provider_type=AMAZON_WEB_SERVICES)
+        with patch.object(AWSReportDownloader, 'download_current_report', return_value=self.file_list):
+            files = downloader.get_current_report()
+            self.assertEqual(len(files), len(self.file_list))
+
+    @patch('masu.external.downloader.aws.aws_report_downloader.AWSReportDownloader.__init__', return_value=None)
+    def test_get_current_report_error(self, fake_downloader):
+        """Test get_current_report function with error."""
+        downloader = ReportDownloader(customer_name='customer name',
+                                      access_credential='mycred',
```
blentz: Needs fake_arn

tests/processor/test_orchestrator.py
```diff
@@ -48,7 +50,25 @@ class OrchestratorTest(MasuTestCase):
...
+                                          'AWS',
+                                          'customer2schema')
+        account3 = CostUsageReportAccount('cred3',
+                                          'billingsource3',
+                                          'customer3',
+                                          'AWS',
+                                          'customer3schema')
+        self.mock_accounts.append(account1)
+        self.mock_accounts.append(account2)
+        self.mock_accounts.append(account3)
```
blentz: This can be more DRY just directly assigning to the mock_accounts array. Or, you can just use `extend()` to append multiple items instead of having three uses of `append()`

Either option is fine by me.

tests/processor/test_orchestrator.py
```diff
@@ -79,4 +99,18 @@ def test_prepare_no_accounts(self, mock_downloader, mock_accounts_accessor):
         orchestrator = Orchestrator()
         reports = orchestrator.prepare()
 
-        self.assertEqual(reports, [])
+        self.assertEqual(len(reports), 0)
```
blentz: Is there a reason to change this? They are functionally equivalent.

IMO, the version without `len()` is more readable.

tests/processor/test_orchestrator.py
```diff
@@ -79,4 +99,18 @@ def test_prepare_no_accounts(self, mock_downloader, mock_accounts_accessor):
...
+
+    @patch('masu.processor.tasks.process.process_report_file', return_value=None)
+    def test_init_with_forced_billing_source(self, mock_task):
+        """Test initializing orchestrator with forced billing source."""
+
+        with patch.object(AccountsAccessor, 'get_accounts', return_value=self.mock_accounts):
+            orchestrator_all = Orchestrator()
+            self.assertEqual(len(orchestrator_all._accounts), len(self.mock_accounts))
```
blentz: Same as before - the list lengths are the same, but are the contents?

masu/external/notification_handler.py
```diff
@@ -0,0 +1,120 @@
...
+        Returns:
+            (Boolean) : True if string is found in header, else False
+
+        """
+        for header in self._headers:
+            if header[0] == value:
+                return True
+        return False
+
+    def get_notified_billing_source(self):
```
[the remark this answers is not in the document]
dccurtis: bad code smell?  I like verbose, explicit names because you really don't know how the ```NotificationHandler``` object will be named when the method is invoked

```
notified_billing_source = handler.get_notified_billing_source()
```

Any suggestions on a better name?

masu/external/notifications/aws/aws_notification_handler.py
```diff
@@ -0,0 +1,174 @@
...
+            (NotificationInterfaceFilter): Is raised when the notification is not for
+                                           a top-level manifest file.
+
+        """
+        s3_bucket_name = None
+        if self._msg_type == 'Notification' and self._body:
+            body_dict = json.loads(self._body)
+            message_dict = json.loads(body_dict['Message'])
+
+            object_key = message_dict['Records'][0]['s3']['object']['key']
```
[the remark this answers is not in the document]
dccurtis: Yeah, I definitely don't like how this is working. I had a comment later down about how this isn't great..  I'll add some more comments explaining that we should get a more deterministic way of extracting the value.

masu/external/notifications/aws/aws_notification_handler.py
```diff
@@ -0,0 +1,174 @@
...
+    """AWS Notification Handler error."""
+
+    pass
+
+
+# pylint: disable=too-few-public-methods
+class AWSNotificationHandler(NotificationInterface):
+    """AWS SNS notification handler."""
+
+    UUID_REGEX = '[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}'
```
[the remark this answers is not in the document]
dccurtis: Good point.  I'll fix this.  I've never seen any uuid from AWS come in capitals, but might as well handle it.  I thought about using the ``uuid``` library but couldn't find a way to detect a uuid in a larger string.  Again, this is pretty much a hack for now, I'll make a comment on it

masu/external/notifications/aws/aws_notification_handler.py
dccurtis: yeah we'll see.  I typically like to keep the dependencies to third party tools to a minimum.  If this proves to be a real pain point then we should look at using something like object path.

masu/external/notifications/aws/aws_notification_handler.py
dccurtis: I don't know if simplified is the right word here, but yeah that will put it on one line :)

tests/external/test_report_downloader.py
dccurtis: what variable is unused?

tests/external/test_report_downloader.py
dccurtis: Sorry. I must be missing something.  fake_arn?

masu/external/__init__.py
```diff
@@ -25,3 +25,24 @@
 
 # Notificaton Definitions
 AWS_SNS_HEADER_MSG_TYPE = 'X-Amz-Sns-Message-Type'
+AWS_SNS_HEADER_TOPIC_ARN = 'X-Amz-Sns-Topic-Arn'
+
+# Valid AWS Regions from https://docs.aws.amazon.com/general/latest/gr/rande.html
+AWS_REGIONS = ['us-east-2',
```
blentz: There's another copy of the aws regions in tests/external/downloader/aws/__init__.py

Let's pick one to keep and remove the other, and update the references.  The version used for testing obviously needs to be static, but the app itself could also use boto's `EC2.Client.describe_regions()` call to fetch a current list, so that we don't have to maintain this as AWS adds/changes their region list.

tests/processor/test_orchestrator.py
dccurtis: ok

tests/processor/test_orchestrator.py
dccurtis: I have no idea.  I think this was a merge-conflict resolution issue or something.  I'll go with (reports, [])

masu/external/__init__.py
dccurtis: yeah I thought about using the describe_regions() function but figured we would need to have some sort of default region to start with to use for that client.  I'll consolidate.

masu/external/notification_handler.py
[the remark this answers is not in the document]
blentz: I don't mind verbose to an extent, but for me there's a point where it gets excessive. Having redundant information in there is across that line for me. (e.g. `foo_module.aws.AWSThingy.get_stuff_from_aws()` is worse, IMO, than `foo_module.aws.Thingy.stuff()`, if that helps.

To me, `.billing_source()` tells us what we need to know - that you're fetching a billing source. Though, `get_billing_source()` follows the conventions you've been using. So, that's okay, too.

masu/external/notifications/aws/aws_notification_handler.py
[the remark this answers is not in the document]
blentz: For this particular case, you could do something like:

```
    for el in object_path.split('/'):
         try:
            my_uuid = uuid.uuid4(el)
         except TypeError:
            continue
```

masu/external/notification_handler.py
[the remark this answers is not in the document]
dccurtis: ok I'll change it

tests/external/test_report_downloader.py
blentz: It's in `tests/external/downloader/aws/__init__.py`. It landed with my async tasks work.

tests/external/test_report_downloader.py
blentz: The `_ =`.  That's an anti-pattern. Please stop using it.

dccurtis: umm ok..  It's nice to know that the function returns something but you don't care what it is.  What's so wrong with that?

dccurtis:
@blentz As discussed offline.  I opened some issues tracking the items you brought up.  I'm going to dismiss the review so we can get this merged to enable QE verification.  Feel free to open other issues on the backlog and we can go over.

PR #82, opened 2018-07-09
set up scheduled tasks

blentz:


blentz:
```
-------------- celery@masu-worker-19-qgpw9 v4.2.0 (windowlicker)
--
  | ---- **** -----
  | --- * ***  * -- Linux-4.16.12-300.fc28.x86_64-x86_64-with-centos-7.5.1804-Core 2018-07-09 17:29:32
  | -- * - **** ---
  | - ** ---------- [config]
  | - ** ---------- .> app:         masu:0x7fbf00aa3a90
  | - ** ---------- .> transport:   amqp://guest:**@rabbitmq.myproject.svc:5672//
  | - ** ---------- .> results:     disabled://
  | - *** --- * --- .> concurrency: 8 (prefork)
  | -- ******* ---- .> task events: ON
  | --- ***** -----
  | -------------- [queues]
  | .> celery           exchange=celery(direct) key=celery
  | .> download         exchange=download(direct) key=download
  | .> process          exchange=process(direct) key=process
  | [tasks]
  | . masu.processor.tasks.check_report_updates
  | . masu.processor.tasks.get_report_files
  | . masu.processor.tasks.process_report_file
  | [2018-07-09 17:29:33,173: INFO/MainProcess] Connected to amqp://guest:**@rabbitmq.myproject.svc:5672//
  | [2018-07-09 17:29:33,192: INFO/MainProcess] mingle: searching for neighbors
  | [2018-07-09 17:29:34,224: INFO/MainProcess] mingle: all alone
  | [2018-07-09 17:29:34,248: INFO/MainProcess] celery@masu-worker-19-qgpw9 ready.
  | [2018-07-09 17:29:35,130: INFO/Beat] beat: Starting...
  | [2018-07-09 17:30:35,240: INFO/Beat] Scheduler: Sending due task check-report-updates (masu.processor.tasks.check_report_updates)
  | [2018-07-09 17:30:35,246: INFO/MainProcess] Received task: masu.processor.tasks.check_report_updates[0ffa09b1-85af-43fc-8787-e5fb520084dc]
  | [2018-07-09 17:30:37,229: INFO/ForkPoolWorker-9] masu.processor.tasks.check_report_updates[0ffa09b1-85af-43fc-8787-e5fb520084dc]: Download task queued for test customer
  | [2018-07-09 17:30:37,242: INFO/ForkPoolWorker-9] Task masu.processor.tasks.check_report_updates[0ffa09b1-85af-43fc-8787-e5fb520084dc] succeeded in 1.993710215203464s: <AsyncResult: ddc3a89c-12cc-477d-a7a1-3b1ab03c852c>
  | [2018-07-09 17:30:37,242: INFO/MainProcess] Received task: masu.processor.tasks.get_report_files[ddc3a89c-12cc-477d-a7a1-3b1ab03c852c]
  | [2018-07-09 17:30:37,244: INFO/ForkPoolWorker-3] masu.processor.tasks.get_report_files[ddc3a89c-12cc-477d-a7a1-3b1ab03c852c]: Downloading report for credential: arn:aws:iam::589173575009:role/CostManagement, source: cost-usage-bucket, customer_name: test customer, provider: AWS
  | [2018-07-09 17:30:37,249: INFO/ForkPoolWorker-3] Found credentials in environment variables.
  | [2018-07-09 17:30:37,268: INFO/ForkPoolWorker-3] Starting new HTTPS connection (1): sts.amazonaws.com
  | [2018-07-09 17:30:45,538: INFO/ForkPoolWorker-3] Starting new HTTPS connection (1): cur.us-east-1.amazonaws.com
  | [2018-07-09 17:30:53,853: INFO/ForkPoolWorker-3] Starting new HTTPS connection (1): cost-usage-bucket.s3.amazonaws.com
  | [2018-07-09 17:31:02,087: WARNING/ForkPoolWorker-3] [2018-07-09 17:31:02,087] INFO in aws_report_downloader: Downloading koku-Manifest.json to /var/tmp/masu/test_customer/aws/koku-Manifest.json
  | [2018-07-09 17:31:02,087: INFO/ForkPoolWorker-3] Downloading koku-Manifest.json to /var/tmp/masu/test_customer/aws/koku-Manifest.json
  | [2018-07-09 17:31:02,095: INFO/ForkPoolWorker-3] Starting new HTTPS connection (2): cost-usage-bucket.s3.amazonaws.com
  | [2018-07-09 17:31:10,811: WARNING/ForkPoolWorker-3] [2018-07-09 17:31:10,811] INFO in aws_report_downloader: Downloading koku-1.csv.gz to /var/tmp/masu/test_customer/aws/koku-1.csv.gz
  | [2018-07-09 17:31:10,811: INFO/ForkPoolWorker-3] Downloading koku-1.csv.gz to /var/tmp/masu/test_customer/aws/koku-1.csv.gz
  | [2018-07-09 17:31:10,813: INFO/ForkPoolWorker-3] Starting new HTTPS connection (3): cost-usage-bucket.s3.amazonaws.com
  | [2018-07-09 17:31:19,285: INFO/ForkPoolWorker-3] Task masu.processor.tasks.get_report_files[ddc3a89c-12cc-477d-a7a1-3b1ab03c852c] succeeded in 42.0415497161448s: None
  | [2018-07-09 17:31:19,286: INFO/MainProcess] Received task: masu.processor.tasks.process_report_file[9d77e23c-447a-4ef0-92e1-f401e68603c7]
  | [2018-07-09 17:31:19,288: INFO/ForkPoolWorker-5] masu.processor.tasks.process_report_file[9d77e23c-447a-4ef0-92e1-f401e68603c7]: Processing Report: schema_name: testcustomer, report_path: /var/tmp/masu/test_customer/aws/koku-1.csv.gz, compression: GZIP
  | [2018-07-09 17:31:20,110: WARNING/ForkPoolWorker-5] [2018-07-09 17:31:20,110] INFO in report_processor: Initialized report processor for file: koku-1.csv.gz and schema: testcustomer
  | [2018-07-09 17:31:20,110: INFO/ForkPoolWorker-5] Initialized report processor for file: koku-1.csv.gz and schema: testcustomer
  | [2018-07-09 17:31:20,193: WARNING/ForkPoolWorker-5] [2018-07-09 17:31:20,193] INFO in report_processor: Completed report processing for file: koku-1.csv.gz and schema: testcustomer
  | [2018-07-09 17:31:20,193: INFO/ForkPoolWorker-5] Completed report processing for file: koku-1.csv.gz and schema: testcustomer
  | [2018-07-09 17:31:20,202: INFO/ForkPoolWorker-5] Task masu.processor.tasks.process_report_file[9d77e23c-447a-4ef0-92e1-f401e68603c7] succeeded in 0.9150578030385077s: None
```

masu/processor/tasks.py
```diff
@@ -101,3 +103,63 @@ def _get_report_files(customer_name,
...
+    stats_recorder.log_last_completed_datetime()
+    stats_recorder.set_cursor_position(last_cursor_position)
+    stats_recorder.commit()
+
+
+@shared_task(name='masu.processor.tasks.check_report_updates', queue_name='celery')
+def check_report_updates():
+    """Scheduled task to initiate scanning process on a regular interval."""
+    reports = []
+    for account in AccountsAccessor().get_accounts():
```
dccurtis: Any reason to do this outside of the Orchestrator?  I was expecting the Orchestrator to be created and then call ```prepare()``` to queue the file download work.  I need to look a little closer but I'm not seeing ```prepare()``` used anywhere in the proposed polling workflow.  It's only in SNS.

blentz: Import loops prevented me from using Orchestrator.  I tried it about a dozen different ways, with no luck.  It's possible @adberglund can spot what I'm doing wrong (again :sweat_smile: ).

dccurtis: Interesting, I'm not following what the import loop is.  Orchestrator would work like this:

1. Init Orchestrator -> Get all accounts ```AccountsAccessor().get_accounts()```
2. orchestrator.prepare() -> walk over all accounts and call ```get_report_files.delay``` 

This is exactly what is happening here, only the logic is being pulled out of the orchestrator.

blentz: I don't claim to understand it (otherwise I'd fix it, rather than working around it).

All I know is that having Orchestrator in the scheduled task constantly gives `could not import Orchestrator` errors. So, there's something in the chain that isn't quite right.

dccurtis: Yeah I understand.  Ironically, your comments around DRY from my last PR really apply here.  We will have 3 ways to kick off the download/process pipeline:

1. Download Endpoint
2. SNS Notification Endpoint
3. Celery polling task

All three of these should create an orchestrator and call ```prepare()``` to queue the download task.  The download endpoint and the celery polling task should do exactly the same thing.  They will create the orchestrator without any parameters to make sure it's getting the full list of accounts.  The SNS notification endpoint is the exception.  That takes a specific bucket and the orchestrator will behave identically as the other case only it has a single account on the list.

This flow is a critical path for masu and it's important that all three paths take exactly the same code path from orchestrator init-time onward.

Lets work through the issue with how the tasks are setup and get this straightened out.  Have you tried just making a new task that runs every X minutes/hours and it does only this:

```
@task_periodic(...)
def poll_for_reports():
    Orchestrator().prepare()
```

masu/celery/tasks.py
```diff
@@ -1,36 +0,0 @@
...
-
-from masu.config import Config
-from masu.processor.orchestrator import Orchestrator
-
-LOG = get_task_logger(__name__)
-
-
-# TODO: Get periodic test to work
-@periodic_task(run_every=Config.REPORT_CHECK_INTERVAL)
-def check_report_updates():
```
dccurtis: This is what I was talking about from my last comment.  Did you try to get this working?  What problems did you run into if so?

masu/processor/tasks.py
adberglund: The import loop is essentially a result of moving all tasks to the same file. The orchestrator is calling one of the tasks, so importing the orchestrator in this file is going to fail. I do agree that if we have an orchestrator as our point of entry, we should use it. The benefit I see to that is that it's a single testable thing, and then we can just call that thing from a few different places (the scheduled polling task, and from the SNS HTTP endpoint) and we have the contract in place that the unit we're calling into is tested. 

To avoid a workaround, we either need separate files for all tasks, or just to move the periodic task to a tasks.py file elsewhere. We could define the periodic task in the celery subpackage? Since it effectively becomes the only direct way we call any of the tasks defined in the processor subpackage, I kind of like it living with the rest of the celery stuff. I believe that would allow importing the orchestrator.

blentz: @adberglund I was having import issues before moving the code around. So, I don't think it's that.

.coveragerc
```diff
@@ -8,7 +8,7 @@ omit =
   */test_*.py
   masu/gunicorn.py
   masu/celery/worker.py
-  masu/celery/tasks.py
+  masu/processor/tasks.py
```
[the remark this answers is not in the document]
blentz: We're ignoring just the celery tasks themselves. The "real task code" is relocated outside of that file into `_tasks`.

masu/celery/tasks.py
blentz: Yes. That's where I started.

The problem is the same import issues as described above.

masu/processor/tasks.py
blentz: @dccurtis - Well, given that my DRY comments from your last PR _didn't_ block the PR, how about we use the same standard?  This code works, even if it isn't perfect and does carry some tech debt.  We can iterate on it.

.coveragerc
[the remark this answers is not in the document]
adberglund: That's not actually the case, though. We have a condition in the get_report_files task (not the _task) that may or may not be executed. And that's how we're triggering the second task to be executed. If that broke, we wouldn't ever process files, and we'd have no tests (or coverage to expose the lack of tests) to tell us that.

masu/processor/tasks.py
dccurtis: Sorry, but comments/concerns on the structure of the actual code is different from test code (you still never approved it btw, so..).  We need to be careful about what kind of debt we are willing to take even though we are just working on a prototype.

For me, I want to get the tasks straightened out rather then working around the problem.  If someone else wants to approve and we can fix it later, I guess I can live with that.  IMHO this isn't a good trade off for just getting it in, it's central to the design of masu.

masu/processor/tasks.py
```diff
@@ -0,0 +1,104 @@
...
+                                report_source,
+                                provider_type,
+                                report_name)
+
+    # initiate chained async task
+    for report_dict in reports:
+        request = {'schema_name': schema_name,
+                   'report_path': report_dict.get('file'),
+                   'compression': report_dict.get('compression')}
+        process_report_file.delay(**request)
```
adberglund: 67-71 is un-tetsted, and the coverage on it is ignored. It is a very important code block.

masu/external/notifications/aws/aws_notification_handler.py
```diff
@@ -26,6 +26,9 @@
 from masu.external.notifications.notification_interface import (NotificationInterface,
                                                                 NotificationInterfaceFilter)
 
+# the cn endpoints aren't supported by moto, so filter them out
+AWS_REGIONS = list(filter(lambda reg: not reg.startswith('cn-'), AWS_REGIONS))
```
dccurtis: Why is this here?  if there is something that Moto can't handle, shouldn't it be in the test code?

dccurtis: Ah sorry, for some reason I read this as downloader!

blentz: The comment says exactly why it's there.

If we select any of the CN-based regions, moto doesn't sandbox the calls and the test contacts the actual AWS endpoint (and gets a `connection refused` because the endpoints don't respond outside of China), which causes test failures for the wrong reasons.

It's working around an upstream bug. I'm going to look at the moto code and see if there's an easy fix we can push upstream.

dccurtis: OK, now that I see this is the notification handler, the question still stands.  By filtering out 'cn-' doesn't that effectively block our ability to handle notifications from 'cn-' regions?

I see you have already done this in the test code to get around the Moto problem, can we remove this in the production code?

blentz: OH! OH! 

That part is definitely outside of the intended scope. I meant to only add that filter to the tests. Looks like I went too far. 

Thanks for catching it. :+1:

PR #88, opened 2018-07-10
require a well-formed ARN for AWS Auth

blentz:
This is wiring up the AwsArn class into the downloader to ensure that we fail early if the auth creds aren't right.


dccurtis:
You probably already know this but the RoleARN is validated on the koku side when the provider is created.
```
Parameter validation failed:
Invalid length for parameter RoleArn, value: 6, valid range: 20-inf
Traceback (most recent call last):
  File "/Users/curtisd/projects/repos/koku_fork/koku/providers/aws/aws_provider.py", line 52, in _get_sts_access
    RoleSessionName='AccountCreationSession'
  File "/Users/curtisd/py_env/koku/lib/python3.6/site-packages/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/curtisd/py_env/koku/lib/python3.6/site-packages/botocore/client.py", line 586, in _make_api_call
    api_params, operation_model, context=request_context)
  File "/Users/curtisd/py_env/koku/lib/python3.6/site-packages/botocore/client.py", line 621, in _convert_to_request_dict
    api_params, operation_model)
  File "/Users/curtisd/py_env/koku/lib/python3.6/site-packages/botocore/validate.py", line 291, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid length for parameter RoleArn, value: 6, valid range: 20-inf
WARNING "POST /api/v1/providers/ HTTP/1.1" 400 14954
```
Keeping the initializer parameter a common string is a little more flexible but I like the idea of validating this just to be extra safe.  Maybe attempting to convert the string parameter in the initializer to AwsArn and throw the appropriate exception might be cleaner.

As it stands now, the SyntaxError would be raised before AWSReportDownloader even gets instantiated and we would have to import AwsArn wherever we use AWSReportDownloader.  It's sort of a pain to have to deal with import dependencies just to use an object.  If the conversion was done inside of AWSReportDownloader initializer then the user would never have to know about this implementation detail and would be presented with a clear exception message instead.

dccurtis:
Yes, that too.  You know how I like to keep such details encapsulated :).  Lets move it into the AWSDownloader initializer.  Thanks!

masu/external/downloader/aws/aws_report_downloader.py
```diff
@@ -69,7 +69,8 @@ def __init__(self, customer_name, auth_credential, bucket, report_name=None, **k
         self.customer_name = customer_name.replace(' ', '_')
 
         LOG.debug('Connecting to AWS...')
-        session = utils.get_assume_role_session(auth_credential, 'MasuDownloaderSession')
+        session = utils.get_assume_role_session(utils.AwsArn(auth_credential),
```
dccurtis: Could we do something like this:
```
try:
    self._rolearn = utils.AwsArn(auth_credential)
except SyntaxError:
    raise AWSDownloaderError('Incorrect RoleARN format')
session = utils.get_assume_role_session(self._rolearn, 'MasuDownloaderSession')
```
That way when it blows up, the backtrace will be very clear about what went wrong.

dccurtis: ok no biggie.

Issue #68, opened 2018-06-26
Add simple check for in-progress or recent data


chambridge:
Add a simple check to avoid double downloading and processing of provider data.

Check start time > completed time.
Check start time within the current hour.
______
Associated with: 
- https://github.com/project-koku/masu/issues/6
- https://github.com/project-koku/masu/issues/7

PR #91, opened 2018-07-12
remove CostUsageReportAccount object, use a dict instead.

blentz:
This is some preparation work to make the data structure being passed around more consistent. This will help make resolving #68 a bit easier.


masu/processor/_tasks/process.py
```diff
@@ -45,17 +45,19 @@ def _process_report_file(schema_name, report_path, compression):
                                 report_path,
                                 compression)
     LOG.info(log_statement)
-
     file_name = report_path.split('/')[-1]
+
     stats_recorder = ReportStatsDBAccessor(file_name)
+    stats_recorder.log_last_started_datetime()
+    stats_recorder.commit()
```
dccurtis: @adberglund Need your db knowledge here.  Is there any unintended consequences of committing here, updating, and committing again later while maintaining the session created in`ReportStatsDBAccessor`?

adberglund: Pretty certain this is fine. Since the log is the only "dirty" object on the session at commit time, it gets committed and that particular transaction is complete. You can then re-use the session to do new stuff (I'm not certain but it might actually retire the session, throw it back in a pool, and then re-grab it again when you call it again), and commit again when done. SQLAlchemy's session will mark dirty state again and consider it a new transaction.

blentz:
Agreed. IMO, this is something that is easily migrated back to an object.

There's even `collections.UserDict` that we can use as the basis for a dict-like object. (See: https://docs.python.org/3.6/library/collections.html?highlight=userdict#collections.UserDict )

PR #95, opened 2018-07-13
93 celery routing

adberglund:
* Remove the shared_task decorator in favor of assigning tasks specifically to the instantiated celery application. 
* Create a universal celery app--instead of using an app factory
  * Do application-specific config updates in a new update function

Manual testing: 

* The download API endpoint inserts an event into the `download` queue, as expected (instead of the default celery queue).
* The scheduled polling task is using the celery queue, and assigning tasks to the appropriate download, and process queues. 
* flower in OpenShift is happy
* download/process tasks run in OpenShift via the download API.


dccurtis:
I'm outside of my wheelhouse on this one so bear with me :)

So the major changes here are:
1. Creating a single celery app.  Before we were creating one in masu and one for the worker.  Where `Celery()` was effectively called twice.  Now `Celery()` is called once in `masu/celery/__init__.py`.  So the masu service creates a flask app and a celery app.  The worker creates only a flask app, right?  Now both the worker and masu are working with the same instance fo the celery app.

2. `@shared_task` is now `@app.task`.  Because we had two celery apps I can see where the `@shared_task` is needed.  Now that we have a concrete instance of the celery app we can be explicit as to what app we are enqueuing a task for.  

From what I've read on this `@shared_task` is intended for reusable apps where you can't make any assumptions about the host app.  We don't have that situation since our worker will always be tightly coupled to our parent app.

Do I generally get the gist of this?

adberglund:
@dccurtis 

To point 1:I think the answer is sort of? The worker still lives separate from the main masu application, and still spawns it's own app and celery instance, so they are technically still separate. Really what's being done here is replacing the celery app factory function with a more concrete variable assignment to a celery instance. That way it can be imported in where it is needed and that allows us to use the @celery.task flag on tasks.

To point 2: Yes, this is really the fix, the work in `1` was just to facilitate this. I don't know what internally was happening that prevented tasks inside of views from having all the details about a task, but this more direct association between task and celery app fixes that. Yes, the relationship here is pretty tightly coupled so I don't feel we need the shared_task decorator.

PR #96, opened 2018-07-13
implement avoiding duplicate processing tasks (closes #68)

### LINKED ISSUE #68: Add simple check for in-progress or recent data

blentz:


masu/processor/tasks.py
```diff
@@ -69,12 +70,22 @@ def get_report_files(customer_name,
...
+        stats = ReportStatsDBAccessor(report_dict.get('file'), provider_id)
+        last_start = stats.get_last_started_datetime()
+        last_end = stats.get_last_completed_datetime()
+
+        # only queue up a processing job if there's none currently in-progress.
+        if not last_start or not last_end or last_start < last_end:
+            request = {'schema_name': schema_name,
+                       'report_path': report_dict.get('file'),
+                       'compression': report_dict.get('compression')}
+            result = process_report_file.delay(**request)
```
adberglund: I know it's a carry over, but at this point making the request container is superfluous. 

`result = process_report_file.delay(schema_name, report_dict.get('file'), report_dict.get('compression'))` takes less space and is, IMHO, more readable. 

Using `*args` and `**kwargs` is great when you maybe have a function that takes a variable number of args and processes whatever you throw at it. In this case we know the task function has 3 well defined parameters and it's 1 less mental step for a reader to see arguments passed directly into a function, instead of mentally unpacking the dictionary to pass them in. I know in this case the dictionary is right there, so it's really not that much additional effort, but my point is, even that small amount of effort need not be expended.

tests/processor/test_tasks.py
```diff
@@ -109,31 +111,43 @@ class TestProcessorTasks(MasuTestCase):
...
-        account = fake_arn(service='iam', generate_account_id=True)
-        schema_name = self.fake.word()
-        get_report_files(customer_name=self.fake.word(),
-                         authentication=account,
-                         provider_type='AWS',
-                         report_name=self.fake.word(),
-                         schema_name=schema_name,
-                         billing_source=self.fake.word(),
-                         provider_id=random.randint(1,65535))
+        get_report_files(**self.fake_get_report_args)
```
adberglund: I know it's easier to write, and in test code, I don't mind the shortcut, really, but to expand on my point on `**kwargs` and readability, here now the reader has to scroll through the file to find the dictionary that is getting passed in to mentally unpack it into the function arguments. The get_report_files function has well defined parameters, so we're not gaining anything by the the use of dictionary unpacking.

tests/processor/test_tasks.py
```diff
@@ -109,31 +111,43 @@ class TestProcessorTasks(MasuTestCase):
 
     fake = faker.Faker()
 
+    def setUp(self):
```
adberglund: Everything done here is quick enough that it's okay to do it for each test, but I just wanted to point out https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUpClass for future reference. You can do a 1 time setup for the whole class for anything that is okay to share between test cases.

tests/processor/test_tasks.py
```diff
@@ -146,18 +160,84 @@ def test_get_report_files_second_task_not_called(self,
...
+                                               mock_started,
+                                               mock_completed):
+        """
+        Test that the chained task is called when start timestamp is before
+        end timestamp.
+        """
+        mock_process_files.delay = Mock()
+        mock_get_files.return_value = self.fake_reports
+
+        mock_started.return_value = None
```
adberglund: Do we need to be concerned with the cases where something is completed, but was never started, or was started but never completed? I know we said that the latter can be addressed by the next time the task is run.

dccurtis:
Agree with Andrew.  The purpose of filling in provider_id is still not clear to me.  In general, since it is a database-centric field I would prefer it to be populated in the lower database layers some how rather than being supplied in the application code itself.  If it were the provider-uuid I think it would make more sense to be done at this level since a UUID is much more universally usable.

All this being said, I don't have any real problem with doing it.  If it's an opportunistic thing to do, just  to make the table complete, that's fine.  But if it's part of a broader strategy; knowing the big picture of how provider_id will be used later would be good to know.

masu/external/downloader/report_downloader_base.py
```diff
@@ -26,14 +26,13 @@ class ReportDownloaderBase():
     Base object class for downloading cost reports from a cloud provider.
     """
 
-    def __init__(self, download_path=None):
+    def __init__(self, download_path=None, provider_id=None):
```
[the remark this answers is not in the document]
blentz: It's a foreign-key reference that was never implemented. It's not core to solving this problem, so I can just strip it out. I initially thought I would need it to solve the duplicate tasks problem, but it turns out that I don't need it after all.

That said, I'd like to at least see the foreign key reference being set when we insert rows into the ReportStats DB. IMO, it's a bug that it isn't being set.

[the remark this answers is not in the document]
dccurtis: Thanks for confirming.  Another option is to just remove provider_id from the table until we have a need for it.

masu/processor/tasks.py
blentz: Understood. I've been using Python long enough that `kwargs` is pretty normal to me, so I don't find it to be significant cognitive effort. To me, it's very idiomatic.

I'm fine with changing it, if desired. I just need clear direction on whether this is a blocker or not.

tests/processor/test_tasks.py
blentz: In this case, it was more of a DRY issue. We were using the same param values across multiple tests.

Where I'm at right now in my learning a more TDD workflow, I find it really helpful to keep the test function itself as clean as possible, so that I can clearly see what's being tested and what pieces are changing from test to test.

tests/processor/test_tasks.py
blentz: AFAICS, there's no possibility for a "completed but never started" case.

There is the possibility of "started, but error causes no completion". Right now, that is not handled at all. It will require admin intervention - the simplest solution is updating the `last_completed_datetime` filed to `now()`.

tests/processor/test_tasks.py
blentz: Totally agree. IMO, the main reason to do it in `setUp()` versus `setUpClass()` is just so that Faker gives us new random values with each test.

masu/processor/tasks.py
adberglund: I might agree if there were keyword arguments to this function with which to use kwargs. As it stands you *must* be passing in a dictionary with **all** of the arguments to the function or the call will fail. Because "explicit is better than implicity" it makes more sense to clearly call out the arguments. The spirit of kwargs is really around flexibility and optional arguments. Because there are no optional arguments, that flexibility doesn't exist here. The following will fail:

```
request = {'schema_name': schema_name, 'report_path': report_dict.get('file')}
result = process_report_file.delay(**request)
```

I was really just trying to point that out with a comment. This isn't blocking, and it doesn't need to be changed.

tests/processor/test_tasks.py
adberglund: Only asked because the test had handling for those cases.

dccurtis: Even though it's a crazy situation, things do happen and it's always nice to know how new logic will react under these unexpected situations.  Because the main `if` condition in this PR is pretty complicated, having a test to ensure each part through it is covered is a worthy goal.

tests/processor/test_tasks.py
```diff
@@ -146,16 +157,125 @@ def test_get_report_files_second_task_not_called(self,
...
+        mock_completed.return_value = self.yesterday
+
+        get_report_files(**self.fake_get_report_args)
+        mock_process_files.delay.assert_not_called()
+
+    @patch('masu.processor.tasks.ReportStatsDBAccessor.get_last_completed_datetime')
+    @patch('masu.processor.tasks.ReportStatsDBAccessor.get_last_started_datetime')
+    @patch('masu.processor.tasks._get_report_files')
+    @patch('masu.processor.tasks.process_report_file')
+    def test_get_report_files_timestamps_empty(self,
```
adberglund: There are now 4 tests with the same function name. Only 1 of them will register and run.

tests/processor/test_tasks.py
```diff
@@ -146,16 +157,125 @@ def test_get_report_files_second_task_not_called(self,
...
+        mock_started.return_value = None
+        mock_completed.return_value = None
+        get_report_files(**self.fake_get_report_args)
+        mock_process_files.delay.assert_called()
+
+    @patch('masu.processor.tasks.ReportStatsDBAccessor.get_last_completed_datetime')
+    @patch('masu.processor.tasks.ReportStatsDBAccessor.get_last_started_datetime')
+    @patch('masu.processor.tasks._get_report_files')
+    @patch('masu.processor.tasks.process_report_file')
+    def test_get_report_files_timestamps_empty(self,
```
adberglund: This one seems to be a duplicate of the first `test_get_report_files_timestamps_empty` above.

PR #99, opened 2018-07-20
A Blueprint to a View

adberglund:
This was work I did while trying to get flask views to respect celery task routes. It turned out to be separate from the solution, but I believe it's still valuable. Flask has the concept of a blueprint which is a nice way to make sets of views modular. In this case I made a blueprint for our API V1, and any views associated with the blueprint get our `/api/v1` route prefix. 

### Examples of usefulness
* This modularity allows us to do things like assign auth to a blueprint, so that only certain views would require it. 
* API versioning becomes as simple as creating another blueprint and associated the required views
* We could have an external API blueprint, and another for internal service communication with different requirements

### Of Note

* The utility functions for blueprints were created to facilitate keeping our views each in their own file. Flask kind of wants you to make all of your blueprint views in a single file, which makes associating the routes with the view simpler. So if we wanted to convert to that style some of this code could go away.


masu/util/blueprint.py
```diff
@@ -0,0 +1,50 @@
...
+#
+
+"""Blueprint util functions."""
+
+from functools import wraps
+
+
+def add_routes_to_blueprint(blueprint, routes):
+    """Assign url route function configuration to given blueprint."""
+    for _, mod_routes in routes.items():
```
blentz: Minor Nit: This could just be `routes.values()` to eliminate the unused key values.

PR #102, opened 2018-07-20
Enable network access to account information.

dccurtis:
**Overview**
Allow for masu to get cost usage report account information over the network rather than reading from the shared database.  This is a step towards a masu/koku microservice architecture.

New environment variables exist to allow for masu to reach koku for account information.
You can see the default settings here:
```
    # Koku Connectivity
    KOKU_HOST = os.getenv('KOKU_HOST_ADDRESS', 'localhost')
    KOKU_PORT = os.getenv('KOKU_PORT', 8000)
    KOKU_ADMIN_USER = os.getenv('KOKU_ADMIN_USEER', 'admin')
    KOKU_ADMIN_PASS = os.getenv('KOKU_ADMIN_PASS', 'pass')

    # Set method for retreiving CUR accounts. 'db' or 'network'
    ACCOUNT_ACCESS_TYPE = os.getenv('ACCOUNT_ACCESS_TYPE', 'db')
```
In order to switch to network access change ACCOUNT_ACCESS_TYPE to 'network'

**Testing**
Missing some negative test case code coverage.  I will increase this at a later point.  Since this code 
is experimental I'm passing on that for now.

Results:
[provider_network_test.txt](https://github.com/project-koku/masu/files/2215000/provider_network_test.txt)


masu/external/accounts/network/cur_accounts_network.py
```diff
@@ -0,0 +1,114 @@
...
+        This will call koku's token-auth to get the service admin token.
+
+        Args:
+            None
+
+        Returns:
+            (String) : Service Admin token
+
+        """
+        token_auth_url = 'http://{}:{}/api/v1/token-auth/'.format(self.koku_host, self.koku_port)
```
chambridge: We should probably support protocol as an environment variable also, so that we can encrypt the password over HTTPS in the future.

masu/config.py
```diff
@@ -37,6 +37,15 @@ class Config(object):
     DB_HOST = os.getenv('DATABASE_HOST', 'localhost')
     DB_PORT = os.getenv('DATABASE_PORT', 15432)
 
+    # Koku Connectivity
+    KOKU_HOST = os.getenv('KOKU_HOST_ADDRESS', 'localhost')
+    KOKU_PORT = os.getenv('KOKU_PORT', 8000)
+    KOKU_ADMIN_USER = os.getenv('KOKU_ADMIN_USEER', 'admin')
```
chambridge: Do we want to stay consistent with the koku openshift template and use `SERVICE_ADMIN_USER`, `SERVICE_ADMIN_PASSWORD`?

masu/external/accounts/network/cur_accounts_network.py
```diff
@@ -0,0 +1,114 @@
...
+        This will return a list of dicts for the Orchestrator to use to access reports.
+
+        Args:
+            None
+
+        Returns:
+            ([{}]) : A list of dicts
+
+        """
+        providers_url = 'http://{}:{}/api/v1/providers/'.format(self.koku_host, self.koku_port)
```
chambridge: Again, support protocol (http | https).

masu/external/accounts/network/cur_accounts_network.py
```diff
@@ -0,0 +1,114 @@
...
+
+        try:
+            response = requests.get(providers_url, headers=self._build_service_admin_auth_header())
```
chambridge: We should check the response here as you might blow up in the next section if its an error that comes back as HTML and you request the `.json()`.

Would it make sense to provide a helper method for this kind of handling?

masu/external/accounts/network/cur_accounts_network.py
dccurtis: Yes.  I can make a note in the code about this.  There are definitely more security related changes needed to be made here.  @adberglund and I were discussing using OpenShift secrets, etc.   The plan was to just keep things simple right now until we prove out the micro-service changes and then come back and do the security work once we (if we) decide to make micro service the default option.

masu/config.py
dccurtis: Since the two services could presumably live on separate hosts I think these environment variables should not be dependent on each other.   But I might as well make them the same just for a simplified local environment where masu and koku are running on the same host.

masu/external/accounts/network/cur_accounts_network.py
dccurtis: Yeah, I'll make some changes to validate the response before `json()`

masu/external/accounts/network/cur_accounts_network.py
dccurtis: Maybe I'm not understanding the comment but where are koku API paths currently defined in masu?  I prefer being able to set the various components as their own environment variables, that makes it easier to change individual pieces rather than having to manipulate a longer base URL string.  Also, don't forget that this is an experimental change, not enabled by default, so I don't want to get bogged down in these types of details too much..

masu/config.py
```diff
@@ -37,6 +37,15 @@ class Config(object):
...
     DB_PORT = os.getenv('DATABASE_PORT', 15432)
 
+    # Koku Connectivity
+    KOKU_HOST = os.getenv('KOKU_HOST_ADDRESS', 'localhost')
+    KOKU_PORT = os.getenv('KOKU_PORT', 8000)
+    KOKU_ADMIN_USER = os.getenv('SERVICE_ADMIN_USER', 'admin')
+    KOKU_ADMIN_PASS = os.getenv('SERVICE_ADMIN_PASSWORD', 'pass')
+
+    # Set method for retreiving CUR accounts. 'db' or 'network'
+    ACCOUNT_ACCESS_TYPE = os.getenv('ACCOUNT_ACCESS_TYPE', 'db')
```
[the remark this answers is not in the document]
dccurtis: I don't know.  I sort of like being able to easily set default values to the individual components in their own environment variable rather than having one big string to establish the default behavior.

Maybe we should save this for if we end up switching to the microservice architecture and the individual environment variables prove to be too onerous.

masu/external/accounts/network/cur_accounts_network.py
```diff
@@ -0,0 +1,114 @@
...
+                msg = 'Token Auth request return code: {}'.format(str(response.status_code))
+                raise CURAccountsInterfaceError(msg)
+        except requests.exceptions.RequestException as error:
+            raise CURAccountsInterfaceError(str(error))
+
+        response_dict = response.json()
+        token = response_dict.get('token')
+        return token
+
+    def _build_service_admin_auth_header(self):
```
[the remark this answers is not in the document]
dccurtis: What function in `requests` allows for creating the Authorization Token header?  I wasn't able to find one.

masu/external/accounts/cur_accounts_interface.py
```diff
@@ -19,6 +19,12 @@
 from abc import ABC, abstractmethod
 
 
+class CURAccountsInterfaceError(Exception):
```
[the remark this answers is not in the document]
dccurtis: ok.  I still don't quite understand why this strategy but I can move it

masu/external/accounts/network/cur_accounts_network.py
```diff
@@ -0,0 +1,114 @@
...
+        Returns:
+            (String) : Service Admin token
+
+        """
+        token_auth_url = 'http://{}:{}/api/v1/token-auth/'.format(self.koku_host, self.koku_port)
+        data = {'username': self.admin_user, 'password': self.admin_pass}
+
+        try:
+            response = requests.post(token_auth_url, data=data)
+            if response.status_code != 200:
```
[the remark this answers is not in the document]
adberglund: Since this is masu, and we're using Flask, I don't think we should be adding a django rest framework dependency to this repo in order to use its constants. requests returns `200` and I think comparing that directly to `200` is okay.

PR #105, opened 2018-07-24
92 bugfix processor duplicate key

adberglund:
## Summary

* Added a method to the ReportDBAccessor to use the Postgres-specific `ON CONFLICT DO NOTHING` clause for INSERTs.
  * If the row does already exist then the primary id for that row is fetched and returned
* Switched the creation of cost entry, cost entry bill, product, pricing, and reservation rows to use the new insert method.
  * This should eliminate the need to worry about the race condition where a row is committed by another process before the current process can commit
* Inserts to these tables will likely be slower now, but we insert to these tables far less frequently than to the line items table, which is not impacted by this change, so the overall change should not be severe

#### Testing 

* Added unit tests to test the new ON CONFLICT usage
* Ran our `api/v1/download/` endpoint with celery running.


tests/database/test_report_db_accessor.py
```diff
@@ -254,6 +256,62 @@ def test_create_db_object(self):
         for column, value in data.items():
             self.assertEqual(getattr(row, column), value)
 
+    def test_insert_on_conflict_do_nothing_with_conflict(self):
+        """Test that an INSERT succeeds ignoring the conflicting row."""
+        # self.accessor.commit()
```
dccurtis: remove

Issue #20, opened 2018-05-24
Cost Usage Data Expiration (3 months)


chambridge:
## User Story
As service, I want to expire customer cost usage data that is older than 3 months so that our database & table size is limited and performance improved rather than keeping data for all time.

The expiration should be a configurable amount with 3 months as a default, but as we move beyond the first release we can contain more data. 

Data expiration should happen on month boundries, it should not delete data on a rolling 3 months from now basis.

## Impacts
Dev

## Assumptions
We can wake up workers on a regular basis and prune data from the data base

## Implementation Details
When each new month hits we should be able to remove any data that is 4 months or older

## Acceptance Criteria

- [ ] Data is removed on a monthly basis for data that is 4 months or older.

dccurtis:
This functionality was checked in under several PRs (including some tooling and bug fixes).  Reviewing the functional tests done for each PR might be helpful in understanding how to test:

https://github.com/project-koku/masu/pull/109: Internal database related changes (nothing external to verify)

https://github.com/project-koku/masu/pull/116: Synchronously deleting expired data (Added `/api/v1/expired_data` endpoint to view and remove expired data

https://github.com/project-koku/masu/pull/119: Asynchronously deleting expired data (Same as #116 only now the deletion happens in celery workers.  Need to now include `remove_expired` queue if running rabbit workers locally

https://github.com/project-koku/masu/pull/122: Continuation of #119 only now expired data queuing happens once a month.  The day of the month can be configured via an environment variable but defaults to the first of the month at midnight.

https://github.com/project-koku/masu/pull/124: Performance optimization (nothing external changed other than log messages)


Tooling needed to setup environment and verify (historical data load by changing masu system date):
https://github.com/project-koku/masu/pull/106

Bug fixed as a result of this work:
https://github.com/project-koku/masu/pull/110

PR #109, opened 2018-07-27
SQLAlchemy changes needed to remove expired report data. WIP #20

dccurtis:
This change is a part of #20.

**Overview**
Masu will ultimately remove report data that is older then 3 months.  This is the first step in accomplishing this.  It focuses strictly on the data base queries and data removal.

A new object `ReportDBCleaner()` is added to kick off the necessary database actions to purge the database of stale data.  While it is completely generic at this point (by taking any expiration date); higher level logic will be added in a subsequent change to calculate the expiration date and provide it to `ReportDBCleaner()`.

**Testing**
Functional Sanity Test:  Did a simple download and process from an AWS bucket to make sure nothing inadvertently broke in ReportProcessing.

Unit tests cover the basic functionality added.  This change does not have to be verified by QE since is deemed to be internal only with no external hooks.


masu/database/report_db_cleaner.py
```diff
@@ -0,0 +1,61 @@
...
+
+        Returns:
+            (None)
+
+        """
+        bill_objects = self._accessor.get_bill_query_before_date(expired_date)
+        for bill in bill_objects.all():
+            bill_id = bill.id
+            line_item_query = self._accessor.get_lineitem_query_for_billid(bill_id)
+            for line_item in line_item_query.all():
```
blentz: I think the performance is going to be very bad, here.

I think we can reduce at least one level of looping by doing `line_item_query.filter(usage_start > expired_date).delete()` to let the DB do more of the heavy-lifting.

dccurtis: yeah probably :). I just wanted to get enough to where something worked in unit test land for now.

So I was running into issues doing .delete() on the Query().
Might be related to this:

http://docs.sqlalchemy.org/en/latest/orm/query.html

see `Warning Additional Caveats for bulk query deletes` part under the `delete()` section

blentz: Yep. This is likely to be a "performance critical" code segment. 

I'm okay leaving this as-is until we do a more performance-focused iteration.  It's up to you if you want to try to tackle it now or leave it for later. :smile:

dccurtis: oh yeah it most definitely is.  It will surely be iterated on in the next step.  By then I will probably figure out how to do the delete on the query.

masu/database/report_db_accessor.py
```diff
@@ -14,7 +14,7 @@
 #    You should have received a copy of the GNU Affero General Public License
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
-"""Downloader for cost usage reports."""
```
adberglund: 😊

masu/database/report_db_accessor.py
```diff
@@ -288,6 +279,28 @@ def get_current_cost_entry_bill(self, bill_id=None):
...
 
+    def get_bill_query_before_date(self, date):
+        """Get the cost entry bill objects with billing period before provided date."""
+        table_name = AWS_CUR_TABLE_MAP['bill']
+        billing_start = getattr(
+            getattr(self.report_schema, table_name),
+            'billing_period_start'
+        )
+        base_query = self._get_db_obj_query(table_name)
+        cost_entry_bill_query = base_query.filter().filter(date <= billing_start)
```
adberglund: double filter?

masu/database/report_db_accessor.py
```diff
@@ -288,6 +279,28 @@ def get_current_cost_entry_bill(self, bill_id=None):
...
+
+    def get_lineitem_query_for_billid(self, bill_id):
+        """Get the AWS cost entry line item for a given bill query."""
+        table_name = AWS_CUR_TABLE_MAP['line_item']
+        cost_entry_bill_id = getattr(
+            getattr(self.report_schema, table_name),
+            'cost_entry_bill_id'
+        )
+        base_query = self._get_db_obj_query(table_name)
+        line_item_query = base_query.filter().filter(cost_entry_bill_id == bill_id)
```
adberglund: Same question.

masu/database/report_db_cleaner.py
adberglund: We can come back to this but yeah, line by line deletes will be bad. We don't actually need any loops. Basically we want to extract the results of the bill query then use those results for a where clause on the delete. Mostly off the top of my head I think the delete is going to go something like `line_item_table_object.delete().where(line_item_table_object.cost_entry_bill_id == _in(bill_ids)` 

I forget how to use `IN` properly in SQLAlchemy-speak

masu/database/report_db_cleaner.py
```diff
@@ -0,0 +1,61 @@
...
+                LOG.debug('Attempting to remove cost item data for bill id: %s, usage_start: %s',
+                          bill_id, line_item.usage_start)
+                self._accessor.session.delete(line_item)
+            self._accessor.session.delete(bill)
```
adberglund: We should probably also delete the `cost_entry` records that reference the bill_id too.

masu/database/report_db_cleaner.py
dccurtis: Yeah I figured we wouldn't ultimately want loops but I wanted some logging statements for when I start functional tests.  I was planning on coming back to tune this up once I know things are working.

masu/database/report_db_accessor.py
dccurtis: doh, I like it super filtered.

masu/database/report_db_accessor.py
dccurtis: copy copy paste paste

masu/database/report_db_cleaner.py
```diff
@@ -0,0 +1,69 @@
...
+            bill_id = bill.id
+
+            line_item_query = self._accessor.get_lineitem_query_for_billid(bill_id)
+            for line_item in line_item_query.all():
+                LOG.debug('Attempting to remove cost item data for bill id: %s, usage_start: %s',
+                          bill_id, line_item.usage_start)
+                self._accessor.session.delete(line_item)
+
+            cost_entry_query = self._accessor.get_cost_entry_query_for_billid(bill_id)
+            for cost_entry in cost_entry_query.all():
```
dccurtis: I know I'm doubling down on this for now.  But I promise it will be efficient before we are done :)

adberglund: Much less of a problem here, anyway. This max size of this table right now is number_of_hours_in_data_retention_period.

PR #110, opened 2018-07-27
Prepend assemblyID to local cost usage report files. Fixes #107

### LINKED ISSUE #107: Report processing incorrectly processes subsequent months of data
Report processing incorrectly processes subsequent months of data.  The file being downloaded is the same name for each billing period and our current download/process implementation will use the last cursor position of the previous month on the current month.  This will lead to the current month's data being incompletely processed.  It is resuming from last months cursor when it should be starting from 0 since it's a new file.

dccurtis:
**Overview**
This change addresses #107 to ensure that unique file names are in the report-stats database which will allow for subsequent month data processing to be successful.

**Testing**
Functional Test Results:
[masu_107_ut.txt](https://github.com/project-koku/masu/files/2236731/masu_107_ut.txt)


masu/util/common.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Args:
+        source_string (Source): string to locate UUIDs.
+
+    Returns:
+        ([]) List of UUIDs found in the source string
+
+    """
+    uuid_regex = '[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}'
+    found_uuid = re.findall(uuid_regex, source_string.lower())
```
blentz: This is a form of yoda condition. 

It's changing the input to match the regex, rather than constructing the regex to match the input (either with `re.IGNORECASE` or by using `[0-9a-fA-F]`).

They're functionally identical, but the preferred style is to hold the input constant and match it with a well-constructed regex.

dccurtis:
Do you mean this for nise?
https://github.com/project-koku/nise/issues/18

dccurtis:
ok lets go with this one then.  Thanks guys

PR #112, opened 2018-07-27
add celery status output

blentz:
Adds celery status to ApplicationStatus

also:

- quiet test output
- add/tweak healthcheck probes in openshift template
- minor tidying/rearranging of status
- minor tidying up of Makefile

Sample liveness output:

```
curl http://masu-myproject.127.0.0.1.nip.io/api/v1/status/?liveness

{"alive":true}
```

Sample status output (slightly modified for readability)
```
curl http://masu-myproject.127.0.0.1.nip.io/api/v1/status/

{"api_version":"1",
"celery_status":{"celery@masu-worker-52-g5nsn":{"active":0,"clock":28506,"freq":2.0,"loadavg":[1.47,1.83,1.85],"local_received":1532719000.9492073,"pid":34,"processed":5,"sw_ident":"py-celery","sw_sys":"Linux","sw_ver":"4.2.0","timestamp":1532719000.9463189,"type":"worker-heartbeat","utcoffset":0}},
"commit":"53a2fc159336a5766e6ae446a8a9cb205c462356",
"current_datetime":"Fri, 27 Jul 2018 19:16:40 GMT",
"debug":false,
"modules":{<OUTPUT_PRUNED>},
"platform_info":{"machine":"x86_64","node":"masu-43-l9jft","processor":"x86_64","release":"4.17.7-200.fc28.x86_64","system":"Linux","version":"#1 SMP Tue Jul 17 16:28:31 UTC 2018"},
"python_version":"3.6.3 (default, Mar 20 2018, 13:50:41) [GCC 4.8.5 20150623 (Red Hat 4.8.5-16)]"}
```


Makefile
```diff
@@ -50,7 +50,10 @@ unittest:
 	tox -e py36
 
 serve:
-	export FLASK_APP=masu; export FLASK_ENV=development; flask run
+	FLASK_APP=masu \
+	FLASK_ENV=development \
+	MASU_SECRET_KEY='t@@ m4nY 53Cr3tZ' \
```
dccurtis: Yes!

masu/__init__.py
```diff
@@ -46,11 +46,15 @@ def create_app(test_config=None):
...
 
     # Load configs
-    if test_config is None:
-        app.config.from_object('masu.config.Config')
-    else:
+    if test_config:
         app.config.from_mapping(test_config)
 
+        # disable log messages less than CRITICAL when running unit tests.
+        logging.disable(logging.CRITICAL)
```
dccurtis: I actually like having logging during tests.  We have one log now that I've been meaning to look into:

`ResourceWarning: unclosed <ssl.SSLSocket fd=31, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('192.168.1.23', 60938), raddr=('54.239.29.25', 443)>`

This most likely means we are doing something wonky in a test and it should probably be looked at.

masu/api/status.py
```diff
@@ -20,42 +20,87 @@
...
+
+        :returns: dict of celery status
+        """
+        state = celery_app.events.State()
+        events = {}
+
+        def announce_worker_event(event):
+            """Announce worker events."""
+            state.event(event)
+            LOG.debug(f'EVENT: {event}')
```
dccurtis: It looks like this would only be printed when the status endpoint is hit.  Would this ever spam?  If not, could we make it `LOG.info`?

openshift/masu-template.yaml
```diff
@@ -237,10 +237,10 @@ objects:
           livenessProbe:
             failureThreshold: 3
             httpGet:
-              path: /api/v1/status
+              path: /api/v1/status/?liveness
```
dccurtis: This is interesting.  Is the `?` standard for liveliness checks?

masu/api/status.py
blentz: In general, I don't think we'd ever really want to see this output outside of a development environment. So, I guess my only objection to changing it is to ask whether you think we're currently over-using INFO when we should be using DEBUG?

openshift/masu-template.yaml
blentz: The `?` just denotes URI query parameters.  (See: RFC3986 & friends)

I was mostly just being lazy and reusing the existing status endpoint over creating an all-new endpoint just to have it output a simple string.

Typically, liveness probes are supposed to be low overhead to just show "early signs of life" within the app. (e.g. a TCP connect to a port or checking for a particular PID) For a Flask app, we start up fast enough, that I can keep the check as an HTTP check that isn't as heavyweight as the status is becoming. 

Adding the checks for Celery involves round-trip communications with RabbitMQ and the Celery workers. So, just checking `/status` now is too slow to be appropriate for a liveness check.

masu/api/status.py
dccurtis: Mainly for debugging in production.  If it's just a line that prints out when the status endpoint is hit, I'd say make it `info`.

That gets to the next part of your question.  When to use `debug`.  I've only found `debug` useful when it's going to seriously spam your log and the noise it produces makes it harder to get to the root cause of the problem your facing.  The debug info is only valuable for very fine grain detail of whats going on.  Think of my temporary loop in the data expiration PR where I'm printing out every single line item being deleted.  There have been *many* times where I've been debugging a production problem and became very frustrated when `debug` was preventing me from getting the info I really needed :)

masu/__init__.py
dccurtis: ok sounds good.  I just don't want to lose visibility of whats going on in our default settings just so dots print pretty during tests.

masu/api/status.py
blentz: That makes sense. So, in this case, I think DEBUG is still appropriate here. 

The contents of the event dict are exactly the same as what is showing up in the status, with a slight change in the structure (we pop `hostname` off the dict to use it as the status dict's key). 

So, it'll be redundant info to what's already getting logged and printed. The only time we'll care about it is during development.

openshift/masu-template.yaml
dccurtis: ok cool, thanks for confirming.  Yeah I knew about `?` for query, thats what threw me off.  Probably not a big deal, but could we make this a new endpoint instead?

blentz: Can we?  Sure. There's lots of ways to skin this cat. :smiley_cat: 

I think my main question is, what does it buy us?

dccurtis: ha I don't know.  It just is a non-standard use of `?`, but we can keep it.

masu/api/status.py
```diff
@@ -20,42 +20,87 @@
...
+            events[event.pop('hostname')] = event
+
+        with celery_app.connection() as connection:
+            recv = celery_app.events.Receiver(connection, handlers={
+                'worker-offline': announce_worker_event,
```
dccurtis: I couldn't figure out how to get the `worker-offline` messages to print but this might be a good candidate to log as an error level some time down the road.  I tried stopping a worker while `masu` was running and didn't see any log message.  In fact,  I only saw `worker-heartbeat` when hitting the status end point.

Since this change would require further changes I'm fine with doing this at a later time if we think it would be valuable.  At the end of the day it would be great to have an error log message when we lose workers.

blentz: 100% agreed. This isn't a perfect or complete implementation by any means. 

The `recv.capture()` call has the details about what it's really doing - we send a message for all workers to wake up; then we listen for 5 seconds for no more than 5 messages; then we report on what was received. So, as long as some workers respond in the time allowed, this check is satisfied.

The exceptions it catches are _broker_ connection issues, not worker connection issues.

So, while yes, we're configured to receive messages other than heartbeats, right now it's highly unlikely we'll see any of those messages unless a worker goes up/down within that 5 second window that the receiver is listening.

LandoCalrizzian:
VERIFIED. Masu status is displaying the correct info and celery status

PR #116, opened 2018-07-31
Synchronously removing expired report data. WIP #20

dccurtis:
**Overview**
This change will synchronously remove expired report data when issuing a `DEL /api/v1/expired_data` when masu is in Debug mode.  This is to demonstrate that removal of report data is successful given the retention policy.  The default retention policy is 3 months of data and the policy can be configured with the environment variable `MASU_RETAIN_NUM_MONTHS`.

Issuing a `GET /api/v1/expired_data` will "simulate" the delete and provide a list of bills that will be removed should the `DELETE` command be executed.

**Testing**
Unit tests were added as well as the following functional tests:
1. Load May - July data from cost-usage-bucket
2. Hit GET /api/v1/expired_data and verify that no entries are returned with the default 3 month retention policy
3. Set env MASU_RETAIN_NUM_MONTHS to 2 and hit GET /api/v1/expired_data
4.  Hit DEL /api/v1/expired_data and verify data for May is actually removed

**Next Steps**
1. Add a new celery task to run at the beginning of each month to purge expired report data.
2. Optimize the database delete in `ReportDBCleaner().purge_expired_report_data()`

Functional Results:
[sync_remove_data_ut.txt](https://github.com/project-koku/masu/files/2243666/sync_remove_data_ut.txt)


dccurtis:
Don't spend a lot of time reviewing `expired_data.py`.  It's changed quite a bit in the next PR.

masu/database/report_db_cleaner.py
```diff
@@ -55,15 +56,23 @@ def purge_expired_report_data(self, expired_date):
             for line_item in line_item_query.all():
                 LOG.debug('Attempting to remove cost item data for bill id: %s, usage_start: %s',
                           bill_id, line_item.usage_start)
-                self._accessor.session.delete(line_item)
+                if simulate is False:
```
blentz: This can just be `if simulate:`, which is the recommended PEP8 style.

For cases where a value defaults to `None`, it makes sense to check for `if foo_var is not None:` rather than the more general `if foo_var:` because various objects will also evaluate to `False` and it can make sense to see if the value differs from `None`.  

But, in this case, we're checking a boolean value already, which means any value that evaluates to `False` is acceptable. So, we can simplify the conditional.

masu/database/report_db_cleaner.py
```diff
@@ -55,15 +56,23 @@ def purge_expired_report_data(self, expired_date):
...
+                    self._accessor.session.delete(line_item)
 
             cost_entry_query = self._accessor.get_cost_entry_query_for_billid(bill_id)
             for cost_entry in cost_entry_query.all():
                 LOG.debug('Attempting to remove cost entry for bill id: %s, interval_start: %s',
                           bill_id, cost_entry.interval_start)
-                self._accessor.session.delete(cost_entry)
+                if simulate is False:
```
blentz: Another style issue here:   `if simulate:`

masu/database/report_db_cleaner.py
```diff
@@ -55,15 +56,23 @@ def purge_expired_report_data(self, expired_date):
...
+                    self._accessor.session.delete(cost_entry)
 
-            self._accessor.session.delete(bill)
-            self._accessor.session.commit()
+            removed_payer_account_id = bill.payer_account_id
+            removed_billing_period_start = bill.billing_period_start
+            if simulate is False:
```
blentz: `if simulate:` here, too.

tests/processor/test_expired_data_remover.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def test_initializer(self):
+        """Test to init"""
+        remover = ExpiredDataRemover('testcustomer')
+        self.assertEqual(remover._months_to_keep, 3)
+        self.assertIsInstance(remover._expiration_date, datetime)
+
+    def test_calculate_expiration_date(self):
+        """Test that the expiration date is correctly calculated."""
+        date_matrix = [{'current_date':    datetime(year=2018, month=7, day=1),
```
blentz: It's a little unclear what cases this text matrix covers. I'd like to see a code comment describing the intent so that we don't have to guess at it.

masu/database/report_db_cleaner.py
```diff
@@ -37,16 +37,17 @@ def __init__(self, schema):
         self._accessor = ReportDBAccessor(schema,
                                           ReportingCommonDBAccessor().column_map)
 
-    def purge_expired_report_data(self, expired_date):
-        """Remove report data with a billing start period before specified dat.
+    def purge_expired_report_data(self, expired_date, simulate=False):
```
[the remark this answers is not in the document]
dccurtis: Yeah, I went back and forth on this.  Will probably make a change in the next PR or two.

masu/database/report_db_cleaner.py
dccurtis: This whole function is going to go through some big changes once I do the performance optimizations. I'll definitely keep this in mind when I settle on the final version.

tests/processor/test_expired_data_remover.py
dccurtis: This is all about taking sure the calculated expiration date is correct given the current date.  I'll add a comment on next round to clarify.

PR #117, opened 2018-07-31
Update the deployment resource limits within the openshift templates.

chambridge:


openshift/rabbitmq.yaml
```diff
@@ -237,3 +240,18 @@ parameters:
...
+- description: Maximum amount of CPU the app container can use.
+  displayName: CPU Limit
+  name: CPU_LIMIT
+  required: true
+  value: '1'
+- displayName: Volume Capacity
+  description: Volume space available for shared files, e.g. 512Mi, 2Gi
+  name: VOLUME_CAPACITY
+  required: true
+  value: 1Gi
```
dccurtis: no idea if lack of new line is a problem or not but just wanted to mention it in case it is.

PR #121, opened 2018-08-02
add PIPENV_PYPI_MIRROR var to buildconfig

blentz:


Makefile
```diff
@@ -75,8 +75,8 @@ oc-dev-db: oc-create-tags oc-create-dev-db
 oc-dev-all: oc-create-tags oc-create-dev-db oc-create-masu oc-create-rabbitmq
 
 oc-create-tags:
-	oc create istag postgresql:9.6 --from-image=centos/postgresql-96-centos7
-	oc create istag python-36-centos7:latest --from-image=centos/python-36-centos7
+	oc create istag postgresql:9.6 --from-image=centos/postgresql-96-centos7 || true
```
adberglund: What does this do?

blentz: It's not supposed to be in this commit, but it allows `make oc-dev-all` to proceed even if that command fails.

PR #119, opened 2018-07-31
Asynchronously removing expired report data. WIP #20

dccurtis:
**Overview**
This is an asynchronous version of https://github.com/project-koku/masu/pull/116

The only change from the synchronous version is that the `expired_data` endpoint will now enqueue the removal tasks and the celery workers will carry out the removal work.

**Testing**
Unit tests were added as well as the following functional tests:
1. Load May - July data from cost-usage-bucket
2. Hit GET /api/v1/expired_data and verify that no entries are returned with the default 3 month retention policy
3. Set env MASU_RETAIN_NUM_MONTHS to 2 and hit GET /api/v1/expired_data
4.  Hit DEL /api/v1/expired_data and verify data for May is actually removed

**Next Steps**
1. Optimize the database delete in `ReportDBCleaner().purge_expired_report_data()`
2. Create a celery beat schedule to call `Orchestrator().remove_expired_report_data()`

Functional Results:
[async_remove_data_ut.txt](https://github.com/project-koku/masu/files/2247152/async_remove_data_ut.txt)

PR #122, opened 2018-08-02
Removing expired report data on regular monthly schedule

dccurtis:
**Overview**
This is a continuation of PR https://github.com/project-koku/masu/pull/119.  Now This same logic kicked off by the `DEL /api/v1/expired_data` call will be ran in a beat schedule that runs on the first of the month.  At midnight the expired report data will be cleared.

A new environment variable `REMOVE_EXPIRED_REPORT_DATA_ON_DAY` was added so we can adjust what day of the month report data cleaning will take place.  If it is set to 0 then report cleaning will be disabled.

Note: I was having trouble setting `crontab` to run on a specific day_of_month as well as specying hour and minutes.  If I find a solution to that I would much prefer to use crontab to just queue a single task at that specific time of the month.  I will surely update this if I find the answer but for now this seems to work.

**Testing**

**NOTE** It's important to know that the worker doing the cleaning is operating on the worker's host date and time.  That being said to test this you will have to set `REMOVE_EXPIRED_REPORT_DATA_ON_DAY` to the current actual day and not necessarily the one set with `MASU_DATE_OVERRIDE` 

Functional test plan:
1. Set env to day 0 to verify that removal task is not scheduled:
2. Unset the env and verify that the cleaning day is defaulted to the first of the month:
3. Test to make sure the cleaning is done at midnight

Results:
[beat_scheudle_remove_data_ut.txt](https://github.com/project-koku/masu/files/2254756/beat_scheudle_remove_data_ut.txt)


dccurtis:
Theres probably a change needed for the OpenShift template.  I'll do that next but figured this was enough to get out for review.

dccurtis:
What sort of configuration items did you have in mind?  I could only thing of setting the specific date and time.  Was there something else?

blentz:
@dccurtis - In short, it should be an Ops decision on what the purge schedule should be, not a hard-coded setting decided by the project developers. 

So, we'd need basically the full crontab-esque range of options to allow an admin to decide whether the purge happens daily, weekly, monthly, or annually, and the precise timing of it (minute, hour, day, day-of-week, day-of-month).

dccurtis:
oh I see.  Yeah need to resolve the problems I having with `crontab` to do that.  For now just focusing on the user story requirements for pilot to be working with 3 months of data for the defaults.

Issue #85, opened 2018-07-10
Convert AWS region strings to easy filtered search


chambridge:
## User Story
As user, I want to be able to filter cost usage data by region so that can see a subset of cost or usage.

## Impacts
Backend

## Role
User

## Assumptions
- This applies to AWS Cost Usage processing only

## Implementation Details
Currently regions within the cost usage reports may be recorded as the following "US East (N. Virginia)". This doesn't lend it self to easily searched data with the UI and does not line up as simple with the availability zones which are `us-east-1`.  When processing the CSV data we should convert the region data to a shorter/simplified form of `us-east`

## Acceptance Criteria

- [ ] Region data is stored in a simpler format; "US East (N. Virginia)" --> "us-east"

LandoCalrizzian:
VERIFIED. I was able to filter based on region/avail_zone without grouping

PR #132, opened 2018-08-08
add facility for maintaining region mapping

blentz:
This adds a facility to maintain a mapping between "region" and "region_name" strings used by AWS to address issue #85 .

Change summary:
- Adds `utils.aws.region_map` functionality
- Adds `/api/v1/regionmap/` API endpoint that triggers pulling region data and inserting into the DB. 

Important: Only INSERT is currently supported, not UPDATE. So, the current flow for updating an existing table will require manually removing data from the DB before triggering a new insertion.  Due to time-boxing this implementation, I'll leave handling updates for a future iteration.

- Fixes a couple of uncaught exceptions in `/api/v1/status/` that I bumped into. 
- Adds `LOG_LEVEL` env var to the masu DeploymentConfig
- Some fine-tuning on the masu readiness healthcheck probe (value decreased from 90 to 30)

Sample DB data:
```
psql (10.4, server 9.6.5)
Type "help" for help.

koku=# select * from region_mapping;
 id |     region     |        region_name         
----+----------------+----------------------------
 25 | us-east-1      | US East (N. Virginia)
 26 | us-east-2      | US East (Ohio)
 27 | us-west-1      | US West (N. California)
 28 | us-west-2      | US West (Oregon)
 29 | ap-south-1     | Asia Pacific (Mumbai)
 30 | ap-northeast-2 | Asia Pacific (Seoul)
 31 | ap-northeast-3 | Asia Pacific (Osaka-Local)
 32 | ap-southeast-1 | Asia Pacific (Singapore)
 33 | ap-southeast-2 | Asia Pacific (Sydney)
 34 | ap-northeast-1 | Asia Pacific (Tokyo)
 35 | ca-central-1   | Canada (Central)
 36 | cn-north-1     | China (Beijing)
 37 | cn-northwest-1 | China (Ningxia)
 38 | eu-central-1   | EU (Frankfurt)
 39 | eu-west-1      | EU (Ireland)
 40 | eu-west-2      | EU (London)
 41 | eu-west-3      | EU (Paris)
 42 | sa-east-1      | South America (SÃ£o Paulo)
 43 | us-gov-west-1  | AWS GovCloud (US)
(19 rows)
```


masu/util/aws/region_map.py
```diff
@@ -0,0 +1,133 @@
...
+
+import json
+import logging
+import re
+
+import bs4
+import requests
+
+LOG = logging.getLogger(__name__)
+URL = 'https://docs.aws.amazon.com/general/latest/gr/rande.html'
```
chambridge: Would it be worth having an environment variable that we could override in case this page changes/goes away?

blentz: This url hasn't changed in years. So, I'm pretty confident that there's a low probability that it'll change.

If it does change anytime soon, I think we can reevaluate then.

PR #137, opened 2018-08-10
Adding support for report prefix to LocalDownloader

dccurtis:
Local downloader can not get reports generated with a report-prefix.  I know the `_extract_names` is sort of kludgy but it works and I didn't want to spend a ton of time polishing it since its just a test tool.  Getting this change in will let us debug the processing issues I was hitting.

Testing:
Unit tests cover the use case.  I also was able to successfully download the local report from the Open Shift team.


masu/external/downloader/local/local_report_downloader.py
```diff
@@ -56,15 +57,52 @@ def __init__(self, customer_name, auth_credential, bucket, report_name=None, **k
...
+            if not item[2]:
+                if any(re.findall(daterange, date) for date in item[1]):
+                    full_path = item[0]
+                    break
+        directories = full_path[len(bucket):]
+        parts = directories.strip('/').split('/')
+
+        report_name = None
+        report_prefix = None
+        if len(parts) == 2:
```
blentz: This block can be simplified and made clearer, like this:

```
report_name = parts.pop(0)
report_prefix = parts.pop(0) if parts else None
```

dccurtis: We would still need to figure out the length of `parts` since if there's a prefix provided, the `report_name` would get the prefix value since it comes second in the list.  So the only difference would be getting the values out directly (like the current implementation) or popping them off the end.  Am I missing something?

dccurtis: oh on second look, removing the 0 it would probably work.

blentz: Ah. Yeah. Small bug in my example. :sweat_smile:

dccurtis: might be a little more too it to, but standby

PR #149, opened 2018-08-21
Adding available memory and disk usage log statements

dccurtis:
Adding log statements for system resources (memory and disk) each time a processing and download task is queued.

Not 100% sure if I should be getting the disk usage for the `/` mount point, but I think its right.  If anyone has any knowledge on how this should be for a persistent volume pod please let me know.

[mem_disk_logs.txt](https://github.com/project-koku/masu/files/2308113/mem_disk_logs.txt)


masu/processor/_tasks/download.py
```diff
@@ -61,6 +62,9 @@ def _get_report_files(customer_name,
                                 customer_name,
                                 provider_type)
     LOG.info(log_statement)
+    disk = psutil.disk_usage('/')
```
adberglund: Should this be the path to the mounted volume instead?

PR #150, opened 2018-08-22
split masu template up

blentz:
This splits the masu template up into four parts: configmaps & secrets, masu, worker, and flower. 

There are now individual parameter files for each template. When deploying from the CLI, you can specify `oc process TEMPLATE --param-file=FILE` to include the values of the parameter file in the template.

The Makefile has been updated to handle the new templates and parameter files. Additionally, the Makefile tries to avoid creating objects that already exist in the running openshift cluster.


Makefile
```diff
@@ -76,11 +79,18 @@ serve:
 oc-clean: oc-down
 	$(PREFIX) rm -rf $(OC_DATA_DIR)
 
-oc-create-all: oc-create-tags oc-create-db oc-create-rabbitmq oc-create-masu
+oc-create-all: oc-create-tags oc-create-configmaps oc-create-db oc-create-rabbitmq oc-create-masu oc-create-worker oc-create-flower
 
 oc-dev-db: oc-create-tags oc-create-db
 
+oc-create-configmaps:
```
adberglund: Let's split out config maps and secrets so they don't have to be deployed together in case you want to change one or the other.

blentz:
Yeah, let's iterate on this. I'm not completely convinced this is the final disposition of these files. It works, but there's definitely more polishing to do.

PR #151, opened 2018-08-22
Just have a single rabbitmq replica for now

adberglund:


adberglund:
Will tackle this as a param in an upcoming issue

PR #156, opened 2018-08-23
Enable configuring flower authentication. (Closes #25)

### LINKED ISSUE #25: Add Flower to RabbitMQ Template
## User Story
As service-admin, I want to run Flower alongside RabbitMQ so that I can see the health of my RabbitMQ service.

## Impacts
RabbitMQ, OpenShift Template

## Role
service admin

## Assumptions
-
-

## Implementation Details
- This should only impact the RabbitMQ openshift template
- The HTTP interface should be configured as HTTPS and require authentication. Using template-defined secrets is acceptable.
- See: http://docs.celeryproject.org/en/latest/userguide/monitoring.html#flower-real-time-celery-web-monitor

## Acceptance Criteria

- [ ] An instance of Flower is available when the Masu RabbitMQ services are deployed.
- [ ] The Flower instance reports the health of all nodes of the RabbitMQ cluster.

blentz:
This adds support for configuring Flower to use authentication. Two authentication types are supported: basic and oauth2. Two OAuth2 providers are supported: google and github.

All of the auth configuration is stored in the masu secrets. To enable auth, simply set the secrets values.

I've verified that basic auth works as expected.

Also changed:
- Move *.env to *.env.example and ignore *.env to avoid accidentally committing local dev configs
- Flower buildconfig is now dockerStrategy to push in a run script that configures auth
- Flower has it's own Pipfile in `openshift/flower`.


Pipfile
```diff
@@ -10,7 +10,6 @@ flask-sqlalchemy = "==2.3.2"
 psycopg2-binary = ">=2.7.0"
 gunicorn = "==19.8.1"
 celery = "==4.2.0"
-flower = "==0.9.2"
```
chambridge: Could we leave this in for those of us that run flower locally?

openshift/flower/s2i/bin/assemble
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
+
+function is_django_installed() {
```
chambridge: Seems like this isn't necessary. Do we need all the django related functions?

openshift/flower/s2i/bin/run
```diff
@@ -0,0 +1,118 @@
...
+set -e
+
+>&2 echo "$(/usr/bin/env)"
+>&2 echo ""
+>&2 echo "PWD=$(pwd)"
+>&2 echo ""
+>&2 echo "$(ls -al)"
+>&2 echo ""
+
+function is_gunicorn_installed() {
```
chambridge: Seems like this isn't necessary. Do we need all the django/gunicorn related functions?

openshift/flower/s2i/bin/assemble
blentz: This is the generic upstream scripting. I've left it unmodified.

Technically, no we don't need it. But, it's also easier to let it match the upstream script than to spend time stripping out the unused functionality.

Pipfile
blentz: I'm reluctant to leave this in for two reasons:

1. It changes the meaning of the Pipfile away from it documenting the dependencies of Masu. Masu doesn't depend on Flower directly, which is why this is getting removed.

2. While folks do run some elements locally from time to time, I think doing so has caused certain problems to remain unresolved for far too long. People are avoiding running these tools from within an openshift cluster, which means they avoid seeing certain classes of problem until we try to run the whole app in stg. 

I think it's time to rip that band-aid off and insist that we standardize development on running all components within an openshift cluster, so that we can deal with the problems that crop up head-on.

chambridge: I get the argument of documenting dependencies, but this can easily be a *dev-package*.

blentz: I can live with that.

PR #162, opened 2018-08-23
Avoid re-processing of CUR files

dccurtis:
**Overview**
Now that we know that AWS CUR files have a new assembly ID each time a report changes we should never have to re-process the same file again as long as we have a started and completed timestamp.  Additionally we want to continue to block re-queuing processing tasks if one is already in progress.

This change will continue to block in flight processing tasks but will additionally not re-process files.  During my experiments I've noticed that hundreds of tasks end up getting queued with only a handful of accounts.

**Testing**
1.  Use nise to generate data for S3.  Ensure that the report is downloaded/processed.
2. Attempt to download/process the same data again.  Verify that processing is not done.
3. Generate new nise data for the same time period and verify that the new data is processed.

Results:
[masu_no_reprocess_ut.txt](https://github.com/project-koku/masu/files/2316136/masu_no_reprocess_ut.txt)


masu/processor/tasks.py
```diff
@@ -71,23 +71,31 @@ def get_report_files(customer_name,
...
-                       'report_path': report_dict.get('file'),
-                       'compression': report_dict.get('compression')}
-            result = process_report_file.delay(**request)
-            LOG.info('Processing task queued - File: %s, Task ID: %s',
-                     report_dict.get('file'),
-                     str(result))
-        else:
-            LOG.info('Processing task NOT queued. Start time is later than end time! Detail: %s',
-                     str(report_dict))
+        # Skip processing if already in progress.
```
chambridge: Any thought to having this be in a time window or something? I’m concerned we could have a worker die during processing and miss the data. Just something to think about bi suppose worst case is we will process the next update.

dccurtis: Yeah maybe if started_date is saved, worker dies before completed timestamp can be saved, then we would never re-process... 
These timestamps are done here:
```
    stats_recorder = ReportStatsDBAccessor(file_name)
    stats_recorder.log_last_started_datetime()
    stats_recorder.commit()
    processor = ReportProcessor(schema_name=schema_name,
                                report_path=report_path,
                                compression=compression)

    processor.process()
    stats_recorder.log_last_completed_datetime()
    stats_recorder.commit()
    stats_recorder.close_session()
```
I think if we remove the commit before processing the database would roll back the change if something bad happened during processing.  @adberglund would this type of change make logging the start/end timestamp be an atomic operation? @blentz, do you remember the reason for adding this commit before processing?  Is the reason still valid now that we know we will never be re-processing the same file outside of a worker dying type situation?

adberglund: Pretty sure it was also for the SNS + Polling case, where it was possible to pick up the same file, so we wanted to log as soon as possible that we had started processing.

chambridge: I don't see it as a blocker, just we may want to open an issue to think about how we want to handle this edge case.

dccurtis: Yeah, I don't want to let this hold up this change since we are wasting lots of resources as is today.  But I'll look at adding a simple timeout window around this check.  If started time is an hour ago with no completed time, then just re-process

PR #166, opened 2018-08-26
Add resource requests parameters for each existing resource limit parameter

LandoCalrizzian:
To allow a customized deployment based on possible OpenShift quotas, this PR adds resource requests parameters for existing resource limit parameter. Updated the .env.example files and added a .env.example file for rabbitmq

Also, add parameters for specifying the number of replicas for rabbitmq and workers.

Flower
- CELERY_FLOWER_REPLICAS
- CELERY_FLOWER_CPU_REQUEST
- CELERY_FLOWER_MEMORY_REQUEST
- BUILD_MEMORY_REQUEST

Masu
- MASU_CPU_REQUEST
- MASU_MEMORY_REQUEST
- BUILD_MEMORY_REQUEST

RabbitMQ
- RABBITMQ_REPLICAS
- MEMORY_REQUEST
- CPU_REQUEST

Worker
- CELERY_WORKER_REPLICAS
- CELERY_WORKER_MEMORY_REQUEST
- CELERY_WORKER_CPU_REQUEST
- BUILD_MEMORY_REQUEST


openshift/masu.env.example
```diff
@@ -2,7 +2,9 @@ NAME=masu
 NAMESPACE=myproject
 BUILD_MEMORY_LIMIT=1Gi
 DATABASE_SERVICE_NAME=koku-pgsql
-MASU_CPU_LIMIT=1
+MASU_CPU_REQUEST='500m'
```
chambridge: seems to be inconsistent as to whether these are quoted.

openshift/worker.yaml
```diff
@@ -245,16 +253,36 @@ parameters:
...
+  displayName: Celery Worker Memory Request
+  name: CELERY_WORKER_MEMORY_REQUEST
+  required: true
+  value: 1Gi
 - description: Maximum amount of memory the Flower container can use.
-  displayName: Memory Limit
+  displayName: Celery Worker Memory Limit
   name: CELERY_WORKER_MEMORY_LIMIT
   required: true
   value: 1Gi
```
adberglund: Let's get the worker limit to the max 2GB

openshift/worker.yaml
```diff
@@ -245,16 +253,36 @@ parameters:
...
+- description: Number of Celery Workers to create
+  displayName: Celery Worker Replicas
+  name: CELERY_WORKER_REPLICAS
+  required: true
+  value: '1'
+- description: Initial amount of memory the Flower container will request.
+  displayName: Celery Worker Memory Request
+  name: CELERY_WORKER_MEMORY_REQUEST
+  required: true
+  value: 1Gi
```
adberglund: If I counted right, as it stands all our requests add up to 4 GB. I think we'll have problems rolling out new updates in this case. A new build or deploy will trigger and fail because we don't have resources to create a new pod (as the new pod gets created before deleting the old). We should lower all our requests to be a fairly low value, just enough to get the thing started, but use as much of our 8GB of `limit`as we can.

openshift/masu.env.example
LandoCalrizzian: Removed the quotes. I copied the values from the yaml but I was able to successfully deploy without any quotes in the param-file

openshift/worker.yaml
LandoCalrizzian: Updated

openshift/worker.yaml
LandoCalrizzian: Setting the initial memory request to 512Mi

openshift/worker.yaml
```diff
@@ -245,16 +253,36 @@ parameters:
...
 - description: Maximum amount of memory the Flower container can use.
-  displayName: Memory Limit
+  displayName: Celery Worker Memory Limit
   name: CELERY_WORKER_MEMORY_LIMIT
   required: true
-  value: 1Gi
+  value: 2Gi
```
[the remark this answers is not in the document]
LandoCalrizzian: That's a good point. Reverting this change.

openshift/worker.env.example
```diff
@@ -1,8 +1,11 @@
 NAME=masu
 NAMESPACE=myproject
 BUILD_MEMORY_LIMIT=1Gi
+CELERY_WORKER_REPLICAS=1
+CELERY_WORKER_CPU_REQUEST=500m
 CELERY_WORKER_CPU_LIMIT=1
-CELERY_WORKER_MEMORY_LIMIT=1Gi
+CELERY_WORKER_MEMORY_REQUEST=512Mi
+CELERY_WORKER_MEMORY_LIMIT=2Gi
```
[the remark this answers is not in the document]
LandoCalrizzian: I agree with you about the purpose of the environment file and will set the value to whatever you all decide. From my testing, the 1Gi memory limit for the single worker deployment has consistently been a hard blocker for processing anything beyond AWS free tier account charges.

PR #177, opened 2018-09-05
Use celery scheduler crontab for hours and minutes for expired data removal. Closes #123

### LINKED ISSUE #123: Fix remove-expired-data beat schedule crontab
While working on https://github.com/project-koku/masu/pull/122 I ran into some problems using `celery.schedules.crontab`.  I was trying to set `crontab(month_of_day=cleaning_day, hour=0, minute=0)` but the task would not be scheduled when the `hour` parameter was used.  I could successfully set the `minute` parameter with `month_of_day` but using `hour` prevented the task from being queued.

I would like to find out what is happening with this and make the correction so we can rely on `crontab` for all aspects of setting the cleaning schedule.  Once this is fixed then the `hour` and `minute` check can be removed from `masu.celery.tasks.remove_expired_data`

dccurtis:
The celery beat schedule defaults to using UTC time.  So the crontab hour parameter must be specified in UTC time.  An environment variable is added so we can specify the hours and minute parameter.  The environment variable was used to test the change.

Testing:
1. Get the current UTC time and set the expired data time ahead.  Set the removal date to the UTC day number.
2. Run celery beat and verify that the remove-expired-data crontab matches the hour and minute.  Verify that the task is ran.
3.  Unset the clean time and verify that it defaults to midnight (hour=0, minute=0)

Test Results:
[masu_issue_123.txt](https://github.com/project-koku/masu/files/2350917/masu_issue_123.txt)

Closes #123


chambridge:
Quick tangent, but if we had multiple workers (e.g. scaled horizontally), I assume they would all try and start expiring data. Wonder if that would cause any issues or just be a waste of cpu cycles. May just be something to think about.

dccurtis:
Currently we have the beat scheduler running in a worker.  @adberglund mentioned this yesterday that we should separate these two since it's not a celery best practice.  Once we make this split then the expired data removal will be decoupled from the worker(s).

PR #180, opened 2018-09-05
Add ReportProcessor interface to support multiple processors

dccurtis:
Masu currently supports processing for AWS cost and usage reports.  In order to support other providers/report types a new processor will have to be created.  This change creates a ReportProcessor interface as a step towards supporting non-AWS providers.

New processors will go in `masu.processors.providers` and it should implement the `process` method so that it easily fits into the masu architecture.

Overview of the change since it has some large diffs, but most of it was just moving/renaming.
1. Moved and renamed `ReportProcessor` to be `AWSReportProcessor` in the `providers` module.
2. Made `ReportProcessor` be the shim between masu and the processors.  It now takes a new provider argument to select the appropriate processor.
3. Passing `provider_type` from the download task to the processing task.

Testing:
1. Create a provider or both AWS and Local
2. Use masu's download endpoint and verify that both AWS and Local data are downloaded and processed

Test Results:
[masu_processor_interface.txt](https://github.com/project-koku/masu/files/2354549/masu_processor_interface.txt)


dccurtis:
It looks like I moved/renamed the ReportProcessor in a way that wiped out the line change history for ReportProcessor.  I'm going to try and fix this such that the line change git history is maintained.

dccurtis:
New PR https://github.com/project-koku/masu/pull/182 which maintains git history.  Closing this PR

PR #183, opened 2018-09-06
Celery scheduler split

adberglund:
## Summary 
Have the celery beat scheduler run in its own build/deploy/pod. It doesn't need the volume mount, all it does is publish tasks on a schedule.


Makefile
```diff
@@ -143,6 +143,18 @@ oc-create-worker: oc-create-configmap oc-create-secrets
 		-p SOURCE_REPOSITORY_REF=$(shell git rev-parse --abbrev-ref HEAD) \
 	| oc create -f -
 
+oc-create-scheduler: oc-create-configmap oc-create-secrets
```
dccurtis: I haven't tried `oc-create-all` in awhile so I'm not entirely sure if it works or not but, you might as well add this to it.

PR #184, opened 2018-09-07
Store AWS account aliases if available.

dccurtis:
**Objective**
AWS account aliases can be used in place of account numbers.  This change will fetch and store the AWS alias for a provider's RoleARN.

In order for aliases to be programmatically retrieved the account admin must grant `IAM ListAccountAliases` access on account's role.

This is a step towards addressing https://github.com/project-koku/koku/issues/329.

**Implementation Notes**
Currently I am refreshing the account alias database table whenever the orchestrator pulls the list of accounts from the database.  I figured we could adjust this later to be either more or less frequent by refreshing at different stages or refreshing in a celery beat schedule.

**Testing**
1. Verify that AWS IAM Policy is not set on the account.  Add a provider and attempt download/process.  Verify log message states that alias not avaiable.
2. Add the ListAccountAliases policy in the IAM dashboard.  Re-download and verify that the alias is set in the database.
3.  Remove the alias and verify that the database is updated on next download.
4. Add a the account alias back and verify db is updated.

Test Results:
[masu_alias_ut.txt](https://github.com/project-koku/masu/files/2361688/masu_alias_ut.txt)


dccurtis:
See my answer on the koku schema PR

PR #186, opened 2018-09-11
make the linter happy.

blentz:
This should fix the linter issues that are blocking the pyup updates.

Also, I've reworked a couple of the class loaders to be dynamic, based on class naming. It's not ideal, but it works with the existing conventions we're currently using. In the future, I'd like to agree on a standardized format to make this even better.


dccurtis:
I'd prefer to separate out the lint changes from the technology-specific objects (Provider, network/db) change since they appear to be two separate issues.

As far as the dynamic class loading goes.  I don't care too much about the details of how each object is instantiated and agree that it feels like theres some sprawl going on with the different types of provider-specific bits (downloader, processor, account stuff, etc.). I have some ideas about how to consolidate things a bit more if the current structure begins to be unmaintainable.  IMO it's still organized and understandable but can see some benefits in getting all of the provider-specific files in a single folder.  The main thing I want to see maintained is that the provider-specific pieces are "loadable" into the masu-core.  (Orchestrator and tasks should never call any provider-specific code directly).

My personal preference is to have the objects be explicitly loaded in some way since it's a lot easier to understand whats happening by just glancing at the code.  Maybe I'm alone on this, if that's the case I have no problem changing it.  If the majority of the developers find it easier to grok the module instantiation pieces by implicitly relying on file naming then lets do it. :)

blentz:
@dccurtis - All good points.

I'll break this PR up into two pieces so we can focus a little more on each part.

My thinking on dynamic loading is that it centers around a question of what things should be made to be "easy". 

I like doing dynamic loading in applications that expect to be extensible and pluggable. We just document whatever the conventions are, and as long as someone puts the the files in the right place with the right name, the application can pick them up when it restarts. It's not too dissimilar to how Python internally loads modules by doing path searches for files named `__init__.py`

For more static applications where we don't expect a lot of outside contributions and/or don't need quite so much flexibility, then we don't need to go through the trouble of setting up a dynamic module loader.

In Masu's case, I think it will make it slightly easier to add in new providers, because it means not needing to find the right place to register new classes and avoiding having giant if-else clauses that enumerate every single possibility. But, you're correct in seeing that there is a tradeoff in where the complexity is.

blentz:
I've split off the dynamic class loading bits, they're now over in PR #189

masu/processor/providers/aws_report_processor.py
```diff
@@ -30,7 +30,7 @@
 from masu.database.report_db_accessor import ReportDBAccessor
 from masu.database.reporting_common_db_accessor import ReportingCommonDBAccessor
 from masu.exceptions import MasuProcessingError
-from masu.external import GZIP_COMPRESSED, UNCOMPRESSED
+from masu.external import GZIP_COMPRESSED
```
dccurtis: I sort of like keeping all compression types here just so it's clear how each type is being opened and if a foreign compression type is mistakenly passed in then it would be implicitly opening with the wrong type.  For that case it would simply not open the file and we would be forced to look into what happened.  All that being said, we are in complete control of when this gets called, so it's probably no problem changing it.

dccurtis:
Thanks for the split!  Yeah I agree that some improvements can be made with how we instantiate provider-specific objects.  Providers are not coming often and there are not going to be hundreds of them so I was going to revisit this after we get OCP under control.

masu/processor/providers/aws_report_processor.py
blentz: In general, the "pythonic" way tends to be to favor simplicity over unnecessary explicitness. That's why pylint complains when you do "if-else" instead of just "if with an implied else", like here.

Preferred:
```
if some-thing:
    do_thing()
do_other_thing()
```

Not preferred:
```
if some-thing:
    do_thing()
else:
    do_other_thing()
```

Functionally, they result in the same thing most of the time, and when the results differ it's often because of other assumptions about what the desired behavior is. So, it's forcing your hand here to re-think what you're really wanting this code to do. (As you accurately observed.)

I think in this particular case, if we wanted to raise an error on an unknown compression type, we should do that explicitly, rather than letting that be a side-effect of falling through the if-clause. 

It's technically true that if we don't know the file is a specific compression, we just have to assume it's uncompressed and wait for the parser failure to surface when the Masu worker tries to read the file. Without doing any additional analysis on the file, we simply can't know at this point in the code whether the file is truly readable.

PR #189, opened 2018-09-13
Implement first pass at dynamic class loading.

blentz:
Split the class loader stuff off from PR #186 to allow evaluation of just these bits. There's some discussion over in that PR about this PR. :sweat_smile:


dccurtis:
Sorry if I wasn't clear from the comment in #186.  I'm fine with changing how provider-specific objects are loaded if it's causing an immediate problem.  If it's not then I'd prefer to revisit the general masu architecture once we complete OCP.  Having some experience living with two providers will give us a better understanding of what, if anything should be refactored.

blentz:
Thanks for the comments. I'll close this out for now and we can revisit later.

PR #204, opened 2018-09-17
Load earlier months of data when adding new providers.

dccurtis:
**Overview**
It is desired that multiple months of data is initially pulled into masu/koku when a new provider account is added.  This change will now download and process the current and previous month by default when a new account is added.  New environment variables are added so we can adjust these in production (`INITIAL_INGEST_OVERRIDE` and `INITIAL_INGEST_NUM_MONTHS`).

**Implementation Highlights**
* The downloader was changed accept an integer parameter to specify how many months of data to download.  (I.e. 2 => current month and previous month)
* Downloader now returns the billing start date as part of the report dictionary.
* The processor needs the billing start date so that it can calculate the summary table data for prior months.
* A new field was added to the provider model to keep track of the initial ingest of report data.

**Testing**
1. Create an S3 bucket with a reporting policy and 3 months of data loaded.  Verify that the initial download/processing includes 2 months of data (current month and previous month)
2. Update the current month's data and re-ingest.  Verify that only the current month is downloaded/processed.
3. Set the INITIAL_INGEST_OVERRIDE env to True and INITIAL_INGEST_NUM_MONTHS to 3.  Verify that the current-2 month is downloaded and processed.  current-1 and current month is not re-processed since they have not had any updates.

Test Results:
[masu_multimonth_initial_ingest_ut.txt](https://github.com/project-koku/masu/files/2390266/masu_multimonth_initial_ingest_ut.txt)


masu/external/downloader/downloader_interface.py
```diff
@@ -24,15 +24,15 @@ class DownloaderInterface(ABC):
...
 
         Implemented by a downloader class.  Must return a list of
         full path of CUR files that are part of the cost usage report.
 
         Args:
-            None
+            (Int) number_of_months - number of prior months to download.
 
         Returns:
             (List) List of full local file paths to CUR files.
```
chambridge: Any thought to just define the method here and let subclasses redefine. I know its an interface, just seems like some copy & paste happening here that could potentially be avoided.

Or maybe put it in `masu/external/report_downloader.py`?

masu/config.py
```diff
@@ -105,3 +105,9 @@ class Config:
 
     # Specify the time of the day for removal of expired report data.
     REMOVE_EXPIRED_REPORT_UTC_TIME = os.getenv('REMOVE_EXPIRED_REPORT_UTC_TIME', '00:00')
+
+    # Specify the number of months (bills) to ingest
+    INITIAL_INGEST_NUM_MONTHS = int(os.getenv('INITIAL_INGEST_NUM_MONTHS', '2'))
+
+    # Override the initial ingest requirement to allow INITIAL_INGEST_NUM_MONTHS
+    INGEST_OVERRIDDE = False if os.getenv('INITIAL_INGEST_OVERRIDE', 'False') == 'False' else True
```
blentz: This should probably be "override".

tests/processor/test_tasks.py
```diff
@@ -192,9 +221,9 @@ def setUp(self):
         self.fake_get_report_args = {'customer_name': self.fake.word(),
                                      'authentication': self.fake_account,
                                      'provider_type': 'AWS',
-                                     'report_name': self.fake.word(),
                                      'schema_name': self.fake.word(),
-                                     'billing_source': self.fake.word()}
+                                     'billing_source': self.fake.word(),
+                                     'provider_uuid': '6e212746-484a-40cd-bba0-09a19d132d64'}
```
blentz: If the uuid doesn't matter for the test, this could easily be `uuid.uuid4()`

masu/config.py
dccurtis: good catch :)

tests/processor/test_tasks.py
dccurtis: ah yeah.  So I'm relying on the test account (`test.sql`) here rather than populating the database in the test case.  Not ideal, but for this purpose seemed ok to me.

PR #205, opened 2018-09-18
Renaming Local provider to AWS-local

dccurtis:
Renaming Local provider to AWS-local to make room for the OCP providers.

Testing: Sanity tested the changes with masu to verify local ingest still works.


dccurtis:
Yeah I was sort of scratching my head at how to tackle the code coverage check. Just disable, merge, and re-enable you think?

PR #208, opened 2018-09-19
Moving around report data and file cleanup to make room for OCP.

dccurtis:
Currently removing temporary report files and expired report data was not done such that AWS was the only option.  This change does some basic renaming and moving of files so that way we can add OCP-specific logic for these cleanup tasks.

Since the diff looks large here's the gist of whats happening:
* ReportDBCleaner -> AWSReportDBCleaner
* ReportProcessor is now responsible for deleting old report files from the temporary volume
* ExpiredDataRemover now takes a provider parameter and picks the appropriate object for removing expired report data.

Testing:
Did basic end-to-end sanity testing covering ingesting and expired data removal by setting the date in the future.


masu/processor/aws/aws_report_processor.py
```diff
@@ -199,6 +200,38 @@ def process(self):
         LOG.info('Completed report processing for file: %s and schema: %s',
                  self._report_name, self._schema_name)
 
+    def remove_temp_cur_files(self, report_path):
```
blentz: Is there some reason we can't use `shutil.rmtree()` to just nuke the entire directory?

PR #209, opened 2018-09-21
Test database schema update and database change for OCP

dccurtis:
This commit is dependent on https://github.com/project-koku/koku/pull/370.  This updates the masu test database schema and also handles the case for database access for bucket to be None since it is not needed for OCP provider.


dccurtis:
https://github.com/project-koku/koku/pull/370 was merged.  This can be approved anytime.

PR #219, opened 2018-10-01
initialize alias with account_id if policy permissions not set

dccurtis:
This change will set the account alias to account id whenever a `ClientError` occurs during alias update.  This is the path that is taken when an AWS account does not have the `IAM ListAccountAliases` policy set.

**Testing**
1. Remove the IAM ListAccountAlias from an account and verify that the account ID is set as the account alis
2. Grant IAM ListAccountAlias permissions and re-download with masu.  Verify that the account alias is set correctly.

Testing Results:
[default_alias_ut.txt](https://github.com/project-koku/masu/files/2434687/default_alias_ut.txt)


tests/util/aws/test_common.py
```diff
@@ -127,6 +127,18 @@ def test_get_account_alias_from_role_arn(self):
...
+        mock_session = mock_get_role_session.return_value
+        mock_client = mock_session.client
+        mock_client.return_value.list_account_aliases.side_effect = ClientError({}, 'Error')
+
+        mock_account_id = '111111111111'
+        role_arn = 'arn:aws:iam::{}:role/CostManagement'.format(mock_account_id)
+
+        account_id, account_alias = utils.get_account_alias_from_role_arn(role_arn, mock_session)
+        self.assertEqual(mock_account_id, account_id)
+        self.assertEqual(mock_account_id, account_alias)
```
blentz: This is effectively the same as `self.assertEqual(account_id, account_alias)`

tests/util/aws/test_common.py
```diff
@@ -127,6 +127,18 @@ def test_get_account_alias_from_role_arn(self):
         self.assertEqual(mock_account_id, account_id)
         self.assertEqual(mock_alias, account_alias)
 
+    @patch('masu.util.aws.common.get_assume_role_session')
+    def test_get_account_alias_from_role_arn_no_policy(self, mock_get_role_session):
+        mock_session = mock_get_role_session.return_value
+        mock_client = mock_session.client
+        mock_client.return_value.list_account_aliases.side_effect = ClientError({}, 'Error')
+
+        mock_account_id = '111111111111'
```
blentz: It would be better to use `Faker().ean(length=13)` here.

PR #222, opened 2018-10-01
OCP Processor database access

dccurtis:
More OCP processor scaffolding to allow masu to access the OCP database tables.

Testing:
Sanity end-to-end download/process test with AWS completed


masu/database/report_db_accessor.py
```diff
@@ -41,10 +41,6 @@ def __init__(self, schema, column_map):
         super().__init__(schema, column_map)
         self._datetime_format = Config.AWS_DATETIME_STR_FORMAT
         self.column_map = column_map
-        self._session = self.get_session()
```
adberglund: Are these just happening in the aws version?

masu/database/__init__.py
```diff
@@ -27,3 +27,12 @@
     'pricing': 'reporting_awscostentrypricing',
     'reservation': 'reporting_awscostentryreservation',
 }
+
+OCP_REPORT_TABLE_MAP = {
+    'report_period': 'reporting_ocpusagereportperiod',
+    'report': 'reporting_ocpusagereport',
+    'usage_report': 'reporting_ocpusagereport',
```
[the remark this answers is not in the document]
dccurtis: good catch, not intentional.  I'll pick one

masu/database/report_db_accessor.py
adberglund: No, they are in the base class now, I just didn't catch that.

PR #224, opened 2018-10-04
OCP report processor

dccurtis:
**Overview**
Report processor for OCP usage reports.  This processor works in the same fashion as the `AWSReportProcessor`.

This will enable us to consume a OCP usage report .csv file at the most basic level.  This means using `OCP-local` provider, it can process a single file in the predetermined `insights_local` directory.

This change is dependent on https://github.com/project-koku/koku/pull/401

**Future work**
* The `remove_temp_cur_files` function can be implemented once we design how multiple report files will be stored in the temporary directory. (This function name should probably change at that time)
* Implement `OCPReportDBCleaner` to remove expired data.
* Once the `cluster_id` concept is well understood and the real `OCPReportDownloader` is complete we should come back and make a more deterministic way of passing cluster_id into the processor.  This implementation extracts it from the report path.
* General code cleanup as we iterate on this.  For example date string conversions.

**Testing**
1. Create an OCP and AWS provider in koku
2.  Add an OpenShift metering usage report file to the local "Insights Upload Service" directory
3. Download/process AWS and OCP data with masu by invoking the download endpoint.
4. Verify database 

**Test Results** 
[masu_ocp_processor_ut.txt](https://github.com/project-koku/masu/files/2447603/masu_ocp_processor_ut.txt)


masu/config.py
```diff
@@ -75,7 +75,7 @@ class Config:
     TMP_DIR = '/var/tmp/masu'
 
     # OCP-local simulates Insights report storage
-    INSIGHTS_LOCAL_REPORT_DIR = f'{TMP_DIR}/insights_local'
+    INSIGHTS_LOCAL_REPORT_DIR = '/var/tmp/insights_local'
```
blentz: Is there some reason this shouldn't be `{TMP_DIR}` ?

masu/config.py
```diff
@@ -84,7 +84,7 @@ class Config:
     REPORT_PROCESSING_BATCH_SIZE = 100000
 
     AWS_DATETIME_STR_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
-    OCP_DATETIME_STR_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
+    OCP_DATETIME_STR_FORMAT = '%Y-%m-%d %H:%M:%S +0000 UTC'
```
blentz: Do we know that this won't clobber valid timezone info?

e.g. If my host node is using UTC+5 for its time zone, will that be honored and passed through to the operator metering reports?

dccurtis: This was just the format that the report had.  @adberglund would know more about how it originated.  We have full control over the report so I could see some of this changing as we go along.

PR #227, opened 2018-10-08
Add account alias to the aggregates table SQL

adberglund:
Add the account alias to the aggregates summary table if it exists. This should help sync calculated totals to aliased accounts.


chambridge:
@adberglund Guessing this needs an update to https://github.com/project-koku/masu/blob/master/tests/sql/test.sql

PR #229, opened 2018-10-09
fix timezone bugs in DateAccessor

blentz:
This fixes issues in DateAccessor related to datetime objects without tzinfo. Now, all datetime instances will have tzinfo set to UTC, if no tzinfo is provided.

Also, the behavior of `MASU_DATE_OVERRIDE` is changed slightly. This PR changes over to using `dateutil.parser` to parse the environment variable. This enables a lot more flexibility in the format. In particular, this lets us specify the environment variable using ISO8601 formatting, which assures that we can pass in tzinfo, if desired.


masu/external/date_accessor.py
```diff
@@ -39,20 +39,23 @@ class DateAccessor():
...
 
     mock_date_time = None
-    date_time_last_accessed = datetime.now()
+    date_time_last_accessed = datetime.now(tz=pytz.UTC)
 
     def __init__(self):
         """Initializer."""
         if Config.MASU_DATE_OVERRIDE and Config.DEBUG and DateAccessor.mock_date_time is None:
-            DateAccessor.mock_date_time = datetime.strptime(Config.MASU_DATE_OVERRIDE,
-                                                            '%Y-%m-%d %H:%M:%S')
```
adberglund: Parser provides flexibility in the face of uncertainty. We know the exact format of our override timestamp. Going by Zen of Python "explicit is better than implicit", I think we should continue to use the explicit `strptime`.

masu/external/date_accessor.py
```diff
@@ -64,20 +67,19 @@ def today(self):
             example: 2018-07-24 15:47:33
 
         """
-        current_date = datetime.today()
+        current_date = datetime.today().replace(tzinfo=pytz.UTC)
```
adberglund: replace doesn't actually change the timezone, so if this is performed with any system timezone other than UTC, it will be inaccurate. 

```
>>> import datetime
>>> import pytz
>>> nowish = datetime.today()
>>> now_utc = datetime.datetime.utcnow()
>>> print(nowish)
2018-10-09 12:47:34.833406
>>> print(now_utc)
2018-10-09 16:47:45.914423
>>> nowish_replaced = nowish.replace(tzinfo=pytz.utc)
>>> print(nowish_replaced)
2018-10-09 12:47:34.833406+00:00
```

For now let's just not touch the `today()` method and just provide a parallel path for the timezone aware values, maybe using separate class attributes for those. It's a crutch, but it gets us through until we address the overarching issue with UTC/timezones.

masu/external/date_accessor.py
blentz: I agree with you on the point of implicit versus explicit. 

However, there's also the principle of "standards-based is better than making something up on your own."

My main motivation for using dateutil is that I know it supports ISO8601 formatting. I know that `datetime.fromisoformat()` isn't available unless we move to Python 3.7. I can look around for a different, explicit option. But, part of my goal is dropping our homegrown formatting in favor of shifting over to the ISO8601 format as what we are going to support.

masu/external/date_accessor.py
blentz: :-1: 

I see no reason adding technical debt is the right answer here, especially when you've already given a better answer - using `utcnow()` instead of `today()`.

tests/external/test_date_accessor.py
```diff
@@ -89,39 +101,58 @@ def test_today_override_override_not_set_debug_false(self):
...
         today = accessor.today()
-
         expected_date = datetime.today()
+
         self.assertEqual(today.year, expected_date.year)
         self.assertEqual(today.month, expected_date.month)
         self.assertEqual(today.day, expected_date.day)
+        self.assertEqual(today.tzinfo, pytz.UTC)
+
+    def test_today_override_with_iso8601(self):
```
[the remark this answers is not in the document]
blentz: I'm leaving the existing tests with the existing format unchanged. 

So, the new test is for showing that ISO8601 format is also supported and working.

PR #234, opened 2018-10-12
OCP Upload Service Integration

dccurtis:
**Overview**
This change introduces Kafka message handling for integration with the Insights upload service.  It listens for two topics `hccm` and `available`.  The `hccm` topic is what the Insights client would use to send OCP usage reports.  The Upload services sends the cost-management app a URL to the file for validation.  The file is validated by successfully extracting it and laying it out into a local directory structure.  Once successfully extracted we respond with a conformation message to the Upload service so the file is avaiable for others to consume.

The local directory structure will temporarly store and organize usage report payloads from OCP clusters.  Masu orchestrator will pick up files from this local staging area at the same intervals as AWS reports ingest.

Listening for Upload service messages is done through a daemon thread running on the masu worker.  This thread has an `asyncio` event loop which will asynchrounsly queue messsages from the `hccm` and `avaiable` topic.  Another `asyncio` task will be pulling messages off the queue for procesing.  Messages from the `hccm` topic are files to be validated.  The handling of these messages involves downloading, extracting, reading the enclosed manifest and filing away into the local directory structure.  The `avaiable` topic is just for demonstration purposes at the moment to show the perminant file link once the hccm msg was validated.

There is a simple packaging script that compresses the report files into a format that masu is expecting.  This involves generating a simple manifest, generating a uuid for the payload, and prepending the file name with the uuid.  This follows a similar pattern of AWS assembly IDs.  The `OCPReportDownloader` will be able to always select the most current report file from the directory based on the most recent manifest.

**Follow on work**
1. Handle multiple files uploads.  The upload service has a cap of 10.5MB.  Any reports larger than this would have to be broken up.
2. De-duplicate the downloaders.  I will open an issue for this work.  I know what to do but didn't want to complicate this PR with those changes.
3. Handle the case where the worker was connected to a kafka server and then the server goes offline.  The kafka msg listener thread does not currently handle this situation.  If the worker never connected to a kafka server then this problem does not occur.
4. Remove `OCP-local`.  This provider type is not needed any more since the `OCP` provider is effectively working from the same local directory.  Populating the local report directory has no dependency on the Upload service being present.
5. Finish unit tests for the kafka msg handling.
6. Implement the temporary file cleanup logic once we settle on the packaging and temporary file handling formats.

**Testing**
1. Start Insights upload service
2.  Start masu service and worker
3. Add OCP Provider in koku
4. Package OCP report into .tar.gz
6. Upload payload to upload service
7. Download endpoint in masu to download/process
8. Add AWS and AWS-local providers to koku and do a sanity check to ensure that all providers are ingested and processed successfully.

**Test Results**
[ocp_report_downloader_ut.txt](https://github.com/project-koku/masu/files/2473180/ocp_report_downloader_ut.txt)


dccurtis:
Sounds good.  Yes I need some sort of usage.  It started as just a convenience for me while developing this but probably time to productize it a little more 👍

PR #241, opened 2018-10-16
Consolidate common downloader logic

### LINKED ISSUE #236: De-duplicate AWS and OCP Report Downloaders
Now that we have two providers, we should de-duplicate the logic between the two providers. Both downloaders are responsible for delivering data for the report processors and consolidating this logic will reduce the chance for coding mistakes.

dccurtis:
Closes #236 

This consolidates some of the report download logic into the top-level `ReportDownloader` class.  This was important to do because there is now a single place populating the report dictionary structures that are passed on to the processor once downloaded.  It also enabled me to lift the lint suppression for `OCPReportDownloader`.

*Testing*
1.  Add an account for AWS, AWS-local, and OCP in koku
2.  Ensure download/processing is successful in masu

*Test Results*
[masu_download_refactor_ut.txt](https://github.com/project-koku/masu/files/2484884/masu_download_refactor_ut.txt)


masu/database/report_db_accessor.py
```diff
@@ -232,7 +232,7 @@ def update_summary_tables(self, provider, start_date, end_date, manifest_id=None
             None
 
         """
-        if provider != 'AWS':
+        if provider not in ('AWS', 'AWS-local'):
```
adberglund: We'll just have to watch who merges first, as my change will invalidate this. My vote is merge yours first, and then I can clean mine.

PR #249, opened 2018-11-01
updated pipfile.lock to fix security concern in requests

adberglund:


Pipfile
```diff
@@ -10,7 +10,7 @@ flask = "==1.0.2"
 psycopg2-binary = ">=2.7.0"
 gunicorn = "==19.8.1"
 celery = "==4.2.0"
-requests = "==2.19.1"
+requests = ">=2.19.1"
```
adberglund:
```suggestion
requests = ">=2.20.0"
```

PR #254, opened 2018-11-06
Calculate OCP charge based on price list

dccurtis:
OCP charges are calculated by multiplying a usage metric (or request metric, but this PR only covers usage) by a cpu/memory rate.

NOTE: I'm using `metric` names `cpu` and `memory` in the `Rate` db model.  There should be some enforcement around this so that koku and masu are in sync.  This will likely come in some of the serializer work for the price list koku API or some other mechanism..

Charges are computed and stored after daily summary is complete.  In addition to that path a new test endpoint is added `/api/v1/update_charge/` to manually invoke the charge update.

**Future Enhancements**
Since the charge update is kicked off in a new celery task we can easily trigger this update whenever the price list koku API gets a new price.  The mechanism would be a similar to that of removing report data in masu when a provider is deleted in koku.

We may want to consider adding a charge for request as well as usage.

**Testing**
1.  Add OCP provider in Koku
2.  Add Rate information directily into the db.
3. Generate OCP data and upload to insights upload service
4.  Verify that the charge info is correct in the database
5. Run update_charge endpoint after updating rates

**Test Results**
[masu_price_list_ut.txt](https://github.com/project-koku/masu/files/2563228/masu_price_list_ut.txt)


dccurtis:
lots of test failures after updating my branch.  Investigating..

tests/processor/test_tasks.py
```diff
@@ -693,3 +715,7 @@ def test_update_summary_tables_ocp_end_date(self):
 
         self.assertEqual(result_start_date, expected_start_date)
         self.assertEqual(result_end_date, expected_end_date)
+
+    def test_update_charge_info_aws(self):
```
adberglund: I think there's missing update charge task tests and/or something didn't get committed.

adberglund: Nope sounds good.

PR #257, opened 2018-11-08
Process ocp tags

adberglund:
## Summary
This adds processing for OCP labels in the OCPReportProcessor. 

## Testing 
I ran the newest version of the OCP operator metering report through the report processor and it loads labels nicely. I also have a unit test case for malformed label string, and the processor doesn't blow up processing a bad label.

## Note
I'd like to merge this after the rate calculation PR as it hits some of the same places, and I will rerun my test.sql  a final time to make sure it covers everything.


masu/processor/ocp/ocp_report_processor.py
```diff
@@ -228,19 +229,48 @@ def _create_usage_report_line_item(self,
...
+        Returns:
+            (dict): The JSON dictionary made from the label string
+
+        """
+        labels = label_string.split('|')
+        label_dict = {}
+
+        for label in labels:
+            try:
+                key, value = label.split(':')
```
chambridge: should we be stripping of `label_` from the key?

adberglund: Yeah, that makes sense.

PR #265, opened 2018-11-19
Fix OCP pod_labels initialization string.

chambridge:


masu/processor/ocp/ocp_report_processor.py
```diff
@@ -229,7 +229,7 @@ def _create_usage_report_line_item(self,
         table_name = OCP_REPORT_TABLE_MAP['line_item']
         data = self._get_data_for_table(row, table_name)
 
-        pod_label_str = '{}'
+        pod_label_str = ''
```
adberglund: Have we run this with something that doesn't have labels? I don't think Postgres will like casting `''` to JSON.

adberglund: Ignore, We're getting back a blank dict anyway.

chambridge: We end up getting `'{}'` because everything passes through `self._process_pod_labels(pod_label_str)`
see this line:
https://github.com/project-koku/masu/blob/master/masu/processor/ocp/ocp_report_processor.py#L243

PR #267, opened 2018-11-20
OCP tier charge updates

dccurtis:
**Overview**
OCP charge calculation can now be applied if tiered rates are included in the Rate db object.  Masu will apply both fixed rate and tiered rate for each line item in the OCP daily summary table (same table used on the Koku API side).

**Testing**
1. Add an OCP Provider and Tiered Rate in koku
2.  Use nise to generate and upload data for cluster.
3.  Download/process from masu
4. Verify charge updates in koku
5. Add a fixed rate to the cpu and memory rates.
6. Refresh rates in masu
7. Verify Koku api

**Test Results**
[masu_charge_tier_ut.txt](https://github.com/project-koku/masu/files/2597514/masu_charge_tier_ut.txt)


chambridge:
Probably unrelated but limit seems to be null for all values and not in the total. We should open a bug and investigate.

dccurtis:
@chambridge Yes, @adberglund said that limit should be available in the latest OCP reports, but nise still doesn't generate limit data.  I'll open that issue.

chambridge:
But `nise` does generate limits. Are they named wrong or something?

https://github.com/project-koku/nise/blob/master/nise/generators/ocp/ocp_generator.py#L183
https://github.com/project-koku/nise/blob/master/nise/generators/ocp/ocp_generator.py#L186

dccurtis:
I'll investigate

PR #271, opened 2018-11-21
OCP CPU and Mem limit schema updates

dccurtis:
Corresponding change to DB schema updates in Koku (https://github.com/project-koku/koku/pull/481)

*Testing*
See end-to-end test results on this PR: https://github.com/project-koku/koku/pull/481


masu/database/sql/reporting_ocpusagelineitem_aggregates.sql
```diff
@@ -7,10 +7,10 @@ CREATE TEMPORARY TABLE reporting_ocpusagelineitem_aggregates_{uuid} AS (
         li.node,
         sum(li.pod_usage_cpu_core_seconds) / 3600 as pod_usage_cpu_core_hours,
         sum(li.pod_request_cpu_core_seconds) / 3600 as pod_request_cpu_core_hours,
-        max(li.pod_limit_cpu_cores) as pod_limit_cpu_cores,
+        max(li.pod_limit_cpu_core_seconds) as pod_limit_cpu_core_hours,
```
adberglund: We can probably get rid of the aggregates file for now, but this has seconds being called hours without unit conversion

masu/database/sql/reporting_ocpusagelineitem_aggregates.sql
```diff
@@ -24,10 +24,10 @@ CREATE TEMPORARY TABLE reporting_ocpusagelineitem_aggregates_{uuid} AS (
         li.node,
         sum(li.pod_usage_cpu_core_seconds) / 3600 as pod_usage_cpu_core_hours,
         sum(li.pod_request_cpu_core_seconds) / 3600 as pod_request_cpu_core_hours,
-        max(li.pod_limit_cpu_cores) as pod_limit_cpu_cores,
+        sum(li.pod_limit_cpu_core_seconds) / 3600 as pod_limit_cpu_core_hours,
         sum(li.pod_usage_memory_byte_seconds / li.total_seconds) * 1e-9 as pod_usage_memory_gigabytes,
         sum(li.pod_request_memory_byte_seconds / li.total_seconds) * 1e-9 as pod_request_memory_gigabytes,
-        max(li.pod_limit_memory_bytes) * 1e-9 as pod_limit_memory_gigabytes
+        sum(li.pod_limit_memory_byte_seconds / li.total_seconds) * 1e-9 as pod_limit_memory_gigabytes
```
adberglund: I think we still want max here. this will be a sum of the limit value in all rows, which is not the limit as set on the pod.

dccurtis:
good points.  I've thought about removing the aggregates table too.  Think this is the time to do it or should we keep it around a little while longer ?

masu/database/sql/reporting_ocpusagelineitem_aggregates.sql
```diff
@@ -7,10 +7,10 @@ CREATE TEMPORARY TABLE reporting_ocpusagelineitem_aggregates_{uuid} AS (
         li.node,
         sum(li.pod_usage_cpu_core_seconds) / 3600 as pod_usage_cpu_core_hours,
         sum(li.pod_request_cpu_core_seconds) / 3600 as pod_request_cpu_core_hours,
-        max(li.pod_limit_cpu_cores) as pod_limit_cpu_cores,
+        max(li.pod_limit_cpu_core_seconds / 3600) as pod_limit_cpu_core_hours,
```
adberglund:
```suggestion
        sum(li.pod_limit_cpu_core_seconds) / 3600 as pod_limit_cpu_core_hours,
```

PR #275, opened 2018-11-30
Catch ReportProcessor tasks in the process celery task

### LINKED ISSUE #216: Error removing temporary processed files
When masu downloads and processed cost usage report files it will remove older files corresponding to older assemblyIDs.  There is currently an exception being thrown on initial ingest.  I have seen this periodically in the past.  However with the latest changes I will be monitoring this to see if it happens on each download/process or if its a one time transient issue that clears itself.

Example:
[https://app.zenhub.com/files/132777214/6d21579a-34f9-42b1-a138-b2cc5fbdd30a/download](https://app.zenhub.com/files/132777214/6d21579a-34f9-42b1-a138-b2cc5fbdd30a/download)

dccurtis:
Closes #216.

When two processing tasks are running simultaneously there was a race condition that was leading to a `FileNotFound` exception when removing old temporary report files.  This exception was not handled and was preventing report summarization from running.

**Test Results**
[masu_race_ut.txt](https://github.com/project-koku/masu/files/2631658/masu_race_ut.txt)


masu/processor/_tasks/process.py
```diff
@@ -71,29 +71,32 @@ def _process_report_file(schema_name, provider, provider_uuid, report_dict):
...
+        provider_accessor = ProviderDBAccessor(provider_uuid=provider_uuid)
+        provider_accessor.setup_complete()
+        provider_accessor.commit()
+        provider_accessor.close_session()
+
+        files = processor.remove_processed_files(path.dirname(report_path))
+        LOG.info('Temporary files removed: %s', str(files))
+
+    except ReportProcessorError as processing_error:
+        LOG.info(str(processing_error))
```
chambridge: Do we want the log level to be warning or error on this?

masu/processor/aws/aws_report_processor.py
```diff
@@ -227,10 +227,13 @@ def remove_temp_cur_files(self, report_path, manifest_id):
...
-                         victim['file'], victim['completed_date'])
-                remove(victim['file'])
-                removed_files.append(victim['file'])
+                try:
+                    LOG.info('Removing %s, completed processing on date %s',
+                             victim['file'], victim['completed_date'])
+                    remove(victim['file'])
+                    removed_files.append(victim['file'])
+                except FileNotFoundError:
+                    LOG.info('Unable to locate file: %s', victim['file'])
```
chambridge: Do we want the log level to be warning or error on this?

masu/processor/_tasks/process.py
dccurtis: oh yeah good point, this should be error level

masu/processor/aws/aws_report_processor.py
dccurtis: this should probably be warning

PR #277, opened 2018-11-30
Ocp node capacity

adberglund:
## Summary
This adds node capacity to the summary SQL calculations for OCP. In addition memory and charge columns were updated to use the `-hours` form. 

This was necessary to for the new current month delta work in the Koku reporting APIs where cluster capacity is needed.


dccurtis:
@adberglund assuming theres a koku schema PR coming 🥇

PR #300, opened 2018-12-17
Add prometheus_flask_exporter to gather metrics for flask app.

chambridge:
Add prometheus_flask_exporter to gather metrics for flask app.

Gathers basic API interaction.

image


openshift/configmap.yaml
```diff
@@ -74,4 +75,9 @@ parameters:
   name: KAFKA_TIMEOUT_SECONDS
   required: true
   value: "30"
+- description: Directory for Prometheus metrics.
+  displayName: Masu Prometheus metrics directory
+  name: PROMETHEUS_DIR
+  required: true
+  value: '/tmp'
```
dccurtis: Just want to double check that we want these metrics going to `/tmp`

chambridge: So we just need an existing directory, we don't need to really retain this data as the Prometheus data scraper asks for data ever 20 seconds or so.

Just picked `/tmp`? You think some other location would be better?

dccurtis: Nope `/tmp` sounds good for that purpose.

PR #305, opened 2018-12-18
Switching OCP data from GB to GiB

### LINKED ISSUE #278: Switch base-10 GB calculations to base-2 GiB calculations for OCP summary data
**Describe the bug**
We made an assumption of base 10 (multiplied bytes by `1e-9` to get GB) but it looks like we have GiB from Openshift and want to use base-2 calculations instead.

dccurtis:
Updating summary calculation to use GiB instead of GB (1GiB ≈ 1.074GB)

*Testing:*
1. Ingest OCP data with master and run Koku API query
2. Switch masu to branch with GiB change and refresh summary data
3. Run same query from 1 and verify that the numbers were updated by the approximation of 1GiB ≈ 1.074GB

*Test Result:*
[masu_gib_convert_ut.txt](https://github.com/project-koku/masu/files/2692445/masu_gib_convert_ut.txt)

Closes #278


dccurtis:
Yeah good point.  I'll touch that up in my next koku PR

PR #307, opened 2018-12-20
Masu docker compose

dccurtis:
See notes and file here: https://github.com/project-koku/koku/pull/539


chambridge:
Same comment as the Koku PR. Looks good. Could we add it to the docs?

PR #319, opened 2019-01-07
Except a socket timeout when checking celery status

adberglund:
## Summary
Except a socket timeout which occurs when rabbit is not present. This will allow the status endpoint (and masu app) to run without rabbit being up


adberglund:
😊

PR #326, opened 2019-01-09
add missing config values to configmap

blentz:
This adds configmap values that Masu supports.


openshift/configmap.yaml
```diff
@@ -34,11 +34,19 @@ objects:
...
-    worker-queues: 'celery,download,process,remove_expired,reporting'
+    worker-queues: "celery,download,process,remove_expired,reporting"
+    schedule-report-checks: "True"
     schedule-check-interval: "60"
     remove-expired-data-on-day: "1"
     remove-expired-data-utc-time: "00:00"
     prometheus-dir: ${PROMETHEUS_DIR}
+    date-override: ""
+    retain-num-months: "3"
+    remove-expired-report-data-on-day: "1"
```
dccurtis: Maybe I'm missing something but is `schedule-report-checks`, `remove-expired-report-data-on-day` and `remove-expired-report-utc-time` used anywhere?

blentz: They're options supported in Masu's `config.py` and Masu is looking to those environment variables for changes. So, ostensibly, a user can set those as environment variables to change the behavior.

This change surfaces them into the openshift config, so that a user doesn't have to RTFS to find them and _then_ manually add them to their deployment config.

PR #334, opened 2019-01-11
Mark aws bill as finalized

adberglund:
## Summary
This will update the bill in the DB with a `finalized_datetime` when AWS data is received that has an invoice_id (the indication that the bill has been finalized).


masu/processor/aws/aws_report_processor.py
```diff
@@ -121,6 +121,7 @@ def line_item_condition_column(self):
         """Create a property with condition to check for line item inserts."""
         return 'invoice_id'
 
+    # pylint:disable=too-many-locals
```
blentz: Can you expound on this comment to explain why so many locals are needed?

This is one of pylint's better warnings that is worth listening to and not disabling.

masu/database/aws_report_db_accessor.py
```diff
@@ -210,3 +210,13 @@ def populate_line_item_aggregate_table(self):
...
+
+    def mark_bill_as_finalized(self, bill_id):
+        """Mark a bill in the database as finalized."""
+        table_name = AWS_CUR_TABLE_MAP['bill']
+
+        bill = self._get_db_obj_query(table_name)\
+            .filter_by(id=bill_id)\
+            .first()
+
+        bill.finalized_datetime = self.date_accessor.today_with_timezone('UTC')
```
blentz: Does it matter that we store two datetimes - a finalized and non-finalized one?

Would it make more sense to have a single datetime and a `finalized`  boolean that tells us the datetime and data shouldn't be modified any further?

adberglund: They're distinct events. The updates can happen at any time, and sometimes more then one time a day. That will eventually be useful to indicate the most recent time we processed data. Finalization happens only 1 time. We use it as a trigger to run an update on the entire billing period (normally we only run 3 day blocks). If we keep the finalization timestamp we can eventually surface when a billing period was finalized. Our summary tables should be considered ephemeral and the update can be run at any time, even after finalization. So we would lose that information with a single timestamp.

masu/processor/aws/aws_report_processor.py
adberglund: I wanted to make the smallest change possible to avoid mucking with critical code. I can break it out into sub-methods.

blentz: I'm okay with just an extra comment. We don't need to refactor anything right now, but I'd like to avoid just a bare pylint disable without any explanation as to why we've disabled it.

PR #358, opened 2019-02-06
Ocp on aws initial cost calculation

adberglund:
## Summary

This is the summary table update for OCP on AWS using our initial simple CPU based cost calculation. 

## Testing
Ran nise for OCP and AWS data using static files with the same resource_id. Then downloaded and ran the updater and confirmed that the summary table is populated.


chambridge:
@adberglund Assuming you have the above static files. I'd suggest creating an `examples` directory in *nise* and dropping them there and claiming this issue also: https://github.com/project-koku/nise/issues/71

Issue #347, opened 2019-01-28
Processing OCP Storage data


chambridge:
## User Story
As user, I want to track my storage usage/request when using OCP so that breakdown my usage/request by cluster, node, project and tag.

## Impacts
API, UI, Docs

## Role
User

## Assumptions
- We will have a second file to process as volumes can exist and not be attached to a pod so pod-level data is not the complete story.
- labels for volumes will contribute to the overall project set of labels (not just pods)

## Implementation Details
- Will need to handle changes to manifest file to read multiple files (to include new storage data)
- Will need to process the raw storage data into tables (similar to current process), aggregate/summarize the data for breakdowns by cluster, node, project, and tags with usage and request data.

## Acceptance Criteria

- [ ] Handles new upload archive with both pod-level cpu/memory and project level storage data.
- [ ] Process data and builds associated tables for storage usage and request raw, aggregate and summary data.
- [ ] Raw storage data is expired after 3 months.

____
Associated with: https://github.com/project-koku/masu/issues/348

PR #359, opened 2019-02-06
Handle multiple .csv files in OCP payload

dccurtis:
**Overview**
The OCP payload which is downloaded from the upload service will now contain a list of files containing both the normal OCP report csv and also second csv file with storage data.

This change makes the connections from Kafka Handler -> OCP Downloader -> OCP Report Processor.

For now the OCP report processor is skipping storage .csv files.  There is some temporary code that will be altered in a follow on PR when we begin processing the storage reports.

**Testing**
1.  Add an OCP provider with cluster id: my-ocp-cluster-1
2.  Generate a nise report csv and package with another storage csv using ocp_csv_package.py script
3. Upload tarball via Insights upload service
4.  Trigger masu download and verify OCP non-storage report data is processed and storage report is ignored

**Test Results**
[masu_ocp_multiple_files_ut.txt](https://github.com/project-koku/masu/files/2838230/masu_ocp_multiple_files_ut.txt)

WIP for #347


dccurtis:
This shouldn't be merged until korekuta in our CI environment is running https://github.com/project-koku/korekuta/pull/44

chambridge:
I don't think we are going to be able to get our OCP 3.10 to return storage reports. What is the breaking issue. If it just `file` -- `files` we could make a simple change to the template and we'd still just pass only the pod usage. Would that work?

dccurtis:
I see.  I was hoping we would be able to update korekuta even if we were not getting storage reports just yet.  Both korekuta and Masu would be handling multiple files.  Breaking change is what you mentioned as well as `files` returning a list instead of a single file string.

I considered making this change backwards compatible but thought it would be easy to just update korekuta for multiple file (even if initially we are only getting a single file) support then merge this.

PR #399, opened 2019-03-15
Redo report finalization

adberglund:
## Summary
This changes AWS report processing to first delete current cost entry line items for the manifest being processed, then doing bulk inserts directly against the line item table. Hashing is removed as there's no need for a uniqueness check when we just re-insert each time. Because resource_id isn't always populated we can't rely on row uniqueness, so we're just nuking and re-inserting. 

## Testing
I used the QA AWS account which was known to cause a problem when ingesting a finalized bill to test. It ingests multiple files and two months of data without error. Our test AWS account works well as well.


tests/processor/aws/test_aws_report_processor.py
```diff
@@ -84,6 +87,24 @@ def setUpClass(cls):
...
+        cls.date_accessor = DateAccessor()
+        billing_start = cls.date_accessor.today_with_timezone('UTC').replace(
+            year=2018,
+            month=6,
+            day=1,
+            hour=0,
+            minute=0,
+            second=0
+        )
+        # datetime.datetime(2018, 6, 1, 0, 0, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None))
```
adberglund:
```suggestion
```

PR #409, opened 2019-04-09
406 add autoscaling

blentz:
Configure autoscaling for #406


chambridge:
So not saying this is bad, but this auto-scales the flask app, right?

The `worker.yaml` is really the key performance change for us; or am I wrong?

blentz:
@chambridge - that's correct. this only scales the masu app - which fires off the initial celery tasks. The worker is a StatefulSet, which can't autoscale. So, we'll need to doing monitoring/alerting around that to know when to scale it up.

Issue #306, opened 2018-12-19
Handle database access and provider delete race condition


dccurtis:
There is a race condition that exists between the masu async worker and when providers are deleted in koku.  Since both services are operating on the same database and there is no locking mechanism in place between the two it is relatively easy to face a situation where masu is in the middle of the download/process pipeline and someone deletes the provider in koku.

This has manifested it self with a few different backtraces in the masu worker log.  Some of them include:

```
[2018-12-19 14:59:40,540: ERROR/ForkPoolWorker-1] Task masu.processor.tasks.get_report_files[5d700cd3-2c7c-4e27-aba2-2d230d6992ab] raised unexpected: AttributeError("'NoneType' object has no attribute 'setup_complete'",)
Traceback (most recent call last):
  File "/opt/app-root/lib/python3.6/site-packages/celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/opt/app-root/src/masu/celery/__init__.py", line 75, in __call__
    return self.run(*args, **kwargs)
  File "/opt/app-root/src/masu/processor/tasks.py", line 72, in get_report_files
    provider_uuid)
  File "/opt/app-root/src/masu/processor/_tasks/download.py", line 60, in _get_report_files
    reports_processed = provider_accessor.get_setup_complete()
  File "/opt/app-root/src/masu/database/provider_db_accessor.py", line 147, in get_setup_complete
    return obj.setup_complete
AttributeError: 'NoneType' object has no attribute 'setup_complete'
[2018-12-19 14:59:40,543: INFO/MainProcess] Received task: masu.processor.tasks.remove_expired_data[879c1796-4371-4d8a-b790-2710756a28fb]  
[2018-12-19 14:59:41,116: ERROR/ForkPoolWorker-1] Task masu.processor.tasks.get_report_files[569f059d-659c-4ed1-9965-d5cb9a6d2187] raised unexpected: AttributeError("'NoneType' object has no attribute 'setup_complete'",)
Traceback (most recent call last):
  File "/opt/app-root/lib/python3.6/site-packages/celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/opt/app-root/src/masu/celery/__init__.py", line 75, in __call__
    return self.run(*args, **kwargs)
  File "/opt/app-root/src/masu/processor/tasks.py", line 72, in get_report_files
    provider_uuid)
[...]

PR #412, opened 2019-04-12
use savepoints in db operations, which should address #306; loads of linting fixes and dedeuplication.

blentz:
This implements using a savepoint by default for most `add()` and `delete()` operations. This required moving redundant methods into the shared parent class. This also simplifies this code and removes some inconsistencies.

Secondarily, while I was in here, I also included a bunch of linting fixes and deduplication of various code paths. In addition to just tidying things up, it also enables us to remove the `skip-file` pylint disables.


masu/database/koku_database_access.py
```diff
@@ -153,3 +159,75 @@ def does_db_entry_exist(self):
...
+        Commit pending database changes.
+
+        Args:
+            None
+        Returns:
+            None
+        """
+        self._session.commit()
+
+    def savepoint(self, func, *args, **kwargs):
```
dccurtis: There is rollback logic in the `KokuDBAccess` when being used as a context manager.  Does `savepoint()` cover anything that the session rollback does not?  Maybe if we were not using a context manager for database accessors?

blentz: See: https://docs.sqlalchemy.org/en/13/orm/session_transaction.html#using-savepoint and https://www.postgresql.org/docs/9.6/sql-savepoint.html

Each savepoint begins a new sub-transaction that can be committed or rolled back. So, it increases the granularity of where the commit/rollback happens.

dccurtis: ok got it.  Yeah I purposefully didn't catch the exceptions in that case because I didn't want to mask any exceptions from our logs for cases other than removing providers while processing.  

I thought I moved all of the database accessors to context managers in the non-test code.

blentz: That could be true. But I think there's an additional benefit in that we can create multiple savepoints, even within a single transaction scope.

So, the context manager will create a savepoint, then each call to `.add()` will create an additional savepoint for cases like this:

```
with MyAccessor as accessor:
    data = fetch_data_somehow()
    for item in data:
        accessor.add(item)
```

Previously, there was one savepoint created at the context manager scope. Now, there's additional savepoints within that scope for each `INSERT` that is within that transaction. So, if a single `INSERT` fails, we'll trap it and log it, but we don't need to rollback the whole transaction.

PR #450, opened 2019-05-21
 Enable cloud watch logging. 

chambridge:
image


masu/util/__init__.py
```diff
@@ -1 +1,51 @@
...
+import os
+
+import watchtower
+from boto3.session import Session
+
+
+def setup_cloudwatch_logging(logger):
+    """Add Cloud Watch log handler if appropriate."""
+    key_id = os.environ.get('CW_AWS_ACCESS_KEY_ID')
+    secret = os.environ.get('CW_AWS_SECRET_ACCESS_KEY')
```
blentz: Please use the standard AWS creds environment variables.

blentz: Is there some reason you aren't using the standard AWS creds environment variables?

Also, is there some reason you're pulling directly from `os.env` and not routing through the Config object?

masu/util/__init__.py
```diff
@@ -1 +1,51 @@
...
+    if not (key_id and secret):
+        logger.info('CloudWatch logging disabled due to missing access key')
+        return
+
+    session = Session(
```
blentz: What happens to Masu if Cloudwatch isn't available?

masu/util/__init__.py
```diff
@@ -1 +1,51 @@
...
+        aws_access_key_id=key_id,
+        aws_secret_access_key=secret,
+        region_name=os.environ.get('CW_AWS_REGION', 'us-east-1'),
+    )
+
+    try:
+        with open('/var/run/secrets/kubernetes.io/serviceaccount/namespace', 'r') as n_file:
+            namespace = n_file.read()
+    except Exception:  # pylint: disable=broad-except
```
blentz: Is there some reason you need this pylint disable?  

Why can't this be formulated in a way to avoid disabling pylint?

masu/celery/__init__.py
```diff
@@ -15,17 +15,30 @@
...
 
 from masu.config import Config
+from masu.util import setup_cloudwatch_logging
 
 
 # pylint: disable=invalid-name, redefined-outer-name
 celery = Celery(__name__, broker=Config.CELERY_BROKER_URL)
 
 
+@after_setup_logger.connect
```
dccurtis: Good to see that this tip was helpful

masu/util/__init__.py
```diff
@@ -1 +1,51 @@
...
+
+import os
+
+import watchtower
+from boto3.session import Session
+
+
+def setup_cloudwatch_logging(logger):
+    """Add Cloud Watch log handler if appropriate."""
+    key_id = os.environ.get('CW_AWS_ACCESS_KEY_ID')
```
dccurtis: Could you add these to the worker.yaml?

blentz:
One other request: Please update your issue description to include more than just a screenshot.

masu/celery/__init__.py
```diff
@@ -15,17 +15,30 @@
...
+def setup_loggers(logger, *args, **kwargs):  # pylint: disable=unused-argument
```
blentz: Please try to reformulate this to not require disabling pylint or please include a comment detailing why it was necessary to disable pylint.

masu/util/__init__.py
chambridge: Yeah. Hit the PR button to soon. Over excited around seeing the data.

masu/util/__init__.py
chambridge: This is a different AWS key/secret than ours. This is from the platforms key/secret to feed data into their Kibana.

I can move it to the config.

masu/util/__init__.py
chambridge: No logs get logged by that log handler into the AWS logging that is picked up by Kibana.

masu/util/__init__.py
chambridge: Sure. Will do. This was just from the platform provided code.

masu/util/__init__.py
blentz: Please also include documentation that discusses this detail.

masu/util/__init__.py
blentz: There's no unit tests that show this and you've provided no functional test output to show that.

Can you please show some test results that demonstrate this code will continue functioning even when boto raises a `ClientError`?

chambridge: Yep, as I mentioned to Doug below. Clicked the PR button sooner than I meant to, while trying to multi-task in a meeting.

masu/celery/__init__.py
chambridge: ok. I'll add a comment about how we don't need to use args, kwargs at this time.

blentz: The code doesn't look like it uses them. Can they be removed?

Issue #444, opened 2019-05-15
Move Kafka handler to new Masu Listener pod


dccurtis:
## User Story
The Kafka handler is currently running as a daemon thread in the masu worker.  In order to scale the workers and ensure temporary file locations are available for processing on the right pod we want to make the following changes:

1. Create a ingest listener pod which will listen to kafka (same as what's happening currently in the worker pod)
2. Instead of simply dropping the file downloaded from the upload service into a temporary staging directory we want to run the line item report processor on it immediately so the line item data is saved in the database.

## Impacts
masu ingest processing

## Assumptions
- Need to clarify whether or not a single listener can be running.  Concern I have: Since we could have multiple listeners subscribed to the same Kafka topic; then both listeners would be downloading/processing the same reports.

## Implementation Details
- It would be nice if the insights local temporary directory was clear after the listener processes the file.  We should keep the insights local directory so we can still side load Nise data and have the poller process pick up and process those changes.

## Acceptance Criteria

- [ ] OpenShift reports will continue to be handled by listening for Kafka messages on the upload service
- [ ] OpenShift reports are processed as soon as the payload is downloaded.

lcouzens:
Verified
Commit: 4ce8120e0b8bf945fcb97c72656cc51465264104

PR #454, opened 2019-05-29
masu-listener: Service for ingesting Kafka based providers

dccurtis:
Addresses #444 
**Overview**
The purpose for this change is to create better infrastructure for ingesting Kafka-based providers.  Today this is applies to OpenShift listening for report updates from the Insights Upload Service.

Enhancements include:
* Ability to scale the number of listeners (as long as it's kept in sync with the number of partitions on the upload service)
* Better resiliency to errors.  With the kafka handler thread being outside of the celery process (as it was on the Masu-worker); we have the ability to not catch certain exceptions.  This would allow for pod restarts via OpenShift.
* Better delineation of Masu `Orchestrator` responsibilities.  Now the `Orchestrator().prepare()` path is reserved strictly for polling based providers (AWS).  This method still works for local OCP ingest used for development and local automated testing.

**Implementation Highlights**
* Kafka handler thread has moved from the Masu-worker and now lives in a new service called masu-listener.
* Report downloads and line item processing is done on the listener.  Once line item processing is complete the data resides in the database.
* Since the report line item data is in the database; the Masu-worker is able to summarize the data.
* masu-listener uses celery and the same RabbitMQ bus to enqueue a summarization task for a Masu-worker to pick up once line item processing was successful.

**Improvement Areas**
* Database accessor changes could be better; but with a merge of Koku/masu on the horizon I didn't spend a lot of time making it pretty.  In the end we will be moving all of this code to Django ORM anyway.
* Kafka thread Asyncio tuning.  There are more opportunities to further parallelize downloads/processes which we will probably do when we start ingest performance testing.  For now I wanted to keep the change as minimal as possible to at least support 2 simultaneously ingests successfully.
* The masu-listener creates the masu app to make the RabbitMQ/celery integration seamless.  This probably isn't necessary and could be refactored if a reason arises.

**Testing**
1. Ingest OCP data via the upload service.  Verify that the listener downloads and processes line item data and that the listener enqueues a worker task to summarize.
2. Ingest OCP data via insights_local.  Verify that the worker treats the OCP account as a polling account and download/processing all happens on the worker.
3. Ingest AWS data and verify it is successful.
4. Ingest OCP data for a provider that does not exist.  Verify that the listener extracts the payload and acknowledges back to the upload service that the payload was successful, but no further procesing is done.
5. Ingest two valid OCP providers via upload services "quickly" (curl command is ran back-to-back) and verify that both providers are successfully processed.

**Test Results**
[masu_listener_ut.txt](https://github.com/project-koku/masu/files/3234109/masu_listener_ut.txt)


masu/external/accounts_accessor.py
```diff
@@ -58,7 +64,22 @@ def _set_source(self):
 
         return None
 
-    def get_accounts(self):
+    @staticmethod
+    def is_polling_account(account):
```
blentz: I'm not sure this makes sense as a staticmethod when the rest of the class expects to be instantiated. This function references none of the class resources, so putting it into this class adds no real value that I can see.

To me, this would be better off as either 1) as a simple utility function in this module outside of the class or 2) In a static class with no instance methods.

(I know we have other staticmethods that are similarly placed into classes. If there's no objections, I'd like to reconcile those at some point, too.)

masu/external/kafka_msg_handler.py
```diff
@@ -194,23 +205,86 @@ def handle_message(msg):
...
+    provider_uuid = utils.get_provider_uuid_from_cluster_id(cluster_id)
+    if provider_uuid:
+        account = get_account(provider_uuid)
+        LOG.info('Processing report for account %s', account)
+
+        reports_to_summarize = get_report_files(**account)
+        LOG.info('Processing complete for account %s', account)
+
+        async_id = summarize_reports.delay(reports_to_summarize)
+        LOG.info('Summarization celery uuid: %s', str(async_id))
```
blentz: Inside this If-statement are three distinct log messages. Multiplied across every provider and every poll cycle, that could add up to a large logfile.

What do you think about doing one of these things?

1. Can we try to reduce this to one logging statement at the bottom of the if-statement.
2. If we really do need log messages for each step, let's look at how many of them can be DEBUG level instead of INFO.

masu/listener/run.py
```diff
@@ -0,0 +1,44 @@
...
+import logging
+
+from masu import create_app
+from masu.celery import celery, update_celery_config
+from masu.config import Config
+from masu.external.kafka_msg_handler import initialize_kafka_handler
+from masu.prometheus_stats import initialize_prometheus_exporter
+
+
+def get_masu_log_level(masu_level):
```
blentz: Can you help me understand why this function is needed?

Where is masu's logging level *NOT* a direct mapping of standard logging levels?  If there's no case where we want INFO to mean DEBUG, then I'm not sure i see why this mapping is necessary.

masu/util/common.py
```diff
@@ -48,3 +54,14 @@ def stringify_json_data(data):
         return str(data)
 
     return data
+
+
+def ingest_method_for_provider(provider):
```
blentz: This feels to me like it should be defined as either a config value or an attribute of the provider class.

Can you help me understand why it makes sense to define these values here?

masu/util/ocp/common.py
```diff
@@ -131,14 +126,35 @@ def get_cluster_id_from_provider(provider_uuid, schema):
...
+    with ProviderAuthDBAccessor(provider_resource_name=cluster_id) as auth_accessor:
+        auth_id = auth_accessor.get_auth_id()
+        if auth_id:
+            with ProviderDBAccessor(auth_id=auth_id) as provider_accessor:
+                provider_uuid = provider_accessor.get_uuid()
+    return provider_uuid
+
+
+def poll_ingest_override_for_provider(provider_uuid):
+    """Determine if provider should be ingested via the poller."""
```
blentz: Can you expand this docstring to be more descriptive, please?

It's not clear to me what the function's purpose is and how it fits into the process.

What is it overriding and why does that need to be overridden? Looking at the code, it doesn't seem to be overriding anything, but instead validating that a path exists.

tests/database/test_provider_db_accessor.py
```diff
@@ -26,54 +26,87 @@
...
         self.assertTrue(accessor.does_db_entry_exist())
         accessor.close_session()
 
+    def test_initializer_auth_id(self):
+        """Test Initializer with authentication database id."""
+        auth_id = self.aws_db_auth_id
+        accessor = ProviderDBAccessor(auth_id=auth_id)
+        self.assertIsNotNone(accessor._session)
+        self.assertTrue(accessor.does_db_entry_exist())
+        accessor.close_session()
```
blentz: Is it necessary to call close_session() at the end of each test?  Shouldn't the `tearDown()` method handle that?

tests/util/ocp/test_common.py
```diff
@@ -76,16 +83,33 @@ def tearDown(self):
...
-    def test_get_multiple_clusters_one_ocp_provider(self):
-        """Test that Exception is raised for multiple cluster IDs on a single OCP provider."""
-        self.creator.create_ocp_report_period(provider_id=self.provider_id)
-        with self.assertRaises(MasuConfigurationError):
-            utils.get_cluster_id_from_provider(self.ocp_test_provider_uuid, self.test_schema)
+    def test_get_provider_uuid_from_cluster_id(self):
+        """Test that the provider uuid is returned for a cluster ID."""
+        cluster_id = self.ocp_provider_resource_name
+        provider_uuid = utils.get_provider_uuid_from_cluster_id(cluster_id)
+        self.assertIsNotNone(provider_uuid)
```
blentz: It would be nice to do some basic type-checking in these unit tests. This should be a valid uuid, not just a non-None value. An empty list or dict would pass this test.

masu/database/provider_auth_db_accessor.py
```diff
@@ -23,16 +23,17 @@
 class ProviderAuthDBAccessor(KokuDBAccess):
     """Class to interact with the koku database for Provider Authentication Data."""
 
-    def __init__(self, auth_id, schema='public'):
+    def __init__(self, auth_id=None, provider_resource_name=None):
```
blentz: It seems unnecessary to drop the `schema` parameter. Unless there's a compelling reason to drop it, I'd suggest just adding on `provider_resource_name` as a new parameter without replacing the `schema` parameter.

adberglund: The compelling reason is that there should never be a case in our current setup where we should allow any schema other than public. Passing in another schema would result in an error.

masu/external/accounts_accessor.py
dccurtis: I'll update the comment to be more descriptive.  But I put it here because it takes an argument that the `AccountsAccessor().get_accounts()` method returns.

masu/external/kafka_msg_handler.py
dccurtis: So just to clarify.  This code is not part of the polling path and it's not for every provider.  This code runs only when a Kafka message comes in with a report to process.

IMO these log statements are important to understand when each step of the processing happens.  Thinking primarily when triaging problems.  I'd prefer to leave this as is.

masu/listener/run.py
dccurtis: Good catch.  It's no longer needed.  Before I integrated the listener with celery (creating the Masu app) I needed to setup logging and this was necessary.  I'l remove it.

masu/util/common.py
dccurtis: Do you have a suggestion of what provider class?  It's just a simple map for the orchestrator to know what provider types should be polled.  I could see it living in the orchestrator itself but figured putting it as a common utility made sense so it could be used in other parts of the code.

masu/util/ocp/common.py
dccurtis: Yes; I'll update the doc string.  This is to support OCP (any LISTENER based provider) data side loading where you can short cut the upload service for development purposes.

tests/database/test_provider_db_accessor.py
dccurtis: Yeah we have a database context manager now that could also be used.  There's several examples in the test code that could be moved to it.  I just didn't bother at this point since it works and my change was big enough...

tests/util/ocp/test_common.py
dccurtis: Sounds good. I'll add that

masu/database/provider_auth_db_accessor.py
dccurtis: Yeah @adberglund got it.  This database table will always be in the public schema

blentz: Okay, in that case the parent class shouldn't have it as a parameter at all, then.

If we're going to remove it, let's remove it from the entire hierarchy.

masu/util/common.py
blentz: I was thinking each Provider class defines its own ingestion method. It could be a class variable or a module-level constant.

masu/database/provider_auth_db_accessor.py
blentz: Okay, in that case the parent class shouldn't have it as a parameter at all, then.

If we're going to remove it, let's remove it from the entire hierarchy.

dccurtis: No.  It's still needed for `KokuDBAccess`.  For example. the report database accessors inherit from `KokuDBAccess` and they supply the appropriate customer tenant schema.

masu/util/common.py
blentz: I was thinking each Provider class defines its own ingestion method. It could be a class variable or a module-level constant.

To me, this looks like configuration data. But, rather than having each provider own it's piece of data, you've centralized all values into a single spot.

The problem is more organizational than technical - it's unclear why this data needed to be centralized when each provider owns other aspects of its own config.

PR #455, opened 2019-05-31
matching on pod name matches for OCP Storage daily table

dccurtis:
The OCP Storage daily summary table was doing a `LEFT JOIN` when determining what nodes a OpenShift volume is running on.  This was causing The node name of NULL to be allowed in the storage daily table.  I believe we should be a `JOIN` here since we only want to match when pod names are aligned between storage and usage.

Because of this the database had NULL values for node and was leading to the 500 seen here: https://github.com/project-koku/koku/issues/867

**Testing**
1. Ingest OCP data with matched pod name for storage and usage
2.  Verify that a group_by[node] && order_by[node] query is successful.
3.  Ingest data with misaligned pod names for storage and usage.
4.  Verify that a group_by[node] && order_by[node] query is successful.

**Test Results**
[koku_issue_867_ut.txt](https://github.com/project-koku/masu/files/3242694/koku_issue_867_ut.txt)


masu/database/sql/reporting_ocpstoragelineitem_daily.sql
```diff
@@ -177,6 +177,8 @@ CREATE TEMPORARY TABLE reporting_ocpstoragelineitem_daily_{uuid} AS (
             ON li.report_id = ur.id
         JOIN reporting_ocpusagereportperiod AS rp
             ON li.report_period_id = rp.id
+        JOIN volume_nodes_{uuid} as uli
```
adberglund: I think we'll drop data if we do a JOIN here. The fact that we did have NULL values making it through means whatever records had a NULL will no longer be in the table. 

I took a single row from the OCP daily storage table that had a null for node and tried to find any record in the raw pod usage table that matched the pod name and there was nothing. That is troubling to me as far as data coming in. But because we don't get node from the volume data and we're assigning it after the fact I think we have to let the nulls through or risk losing usage data for storage.

dccurtis: Yeah lets look into this a little deeper before merging anything.

dccurtis:
Closing in favor of https://github.com/project-koku/koku/pull/883

PR #458, opened 2019-06-03
adding process queue to celery config.

dccurtis:
The process queue was missing from our OpenShift celery config.  The new celery task `summarize_reports` was added to this queue and explains why report summarization has not been running.


dccurtis:
I've already updated the CI and QA environments and see the process queue burning down now.

Issue #462, opened 2019-06-12
masu-listener becomes unusable when line item processing fails.


dccurtis:
The following exception was seen in hccm-qa.  While it's not clear why this database schema issue occurred the larger problem was that the listener appeared to be blocked awaiting on the `process_report` thread to return.  The listener was completely locked and unable to service any subsequent processing events.

The problem is that there is a try/catch block missing around the `run_in_executor` call

```
[2019-06-11 13:20:00,013] INFO in ocp_report_processor: File <_io.TextIOWrapper name='/var/tmp/masu/acct477931/ocp/test3-ocp-delta/20190601-20190701_6e07e100-71cc-42f5-a635-5f8844aae091_openshift_usage_report.0.csv' mode='r' encoding='UTF-8'> opened for processing
[2019-06-11 13:20:05,806] INFO in ocp_report_processor: Saving report rows 0 to 253 for 20190601-20190701_6e07e100-71cc-42f5-a635-5f8844aae091_openshift_usage_report.0.csv
[2019-06-11 13:20:05,807] INFO in ocp_report_processor: Completed report processing for file: /var/tmp/masu/acct477931/ocp/test3-ocp-delta/20190601-20190701_6e07e100-71cc-42f5-a635-5f8844aae091_openshift_usage_report.0.csv and schema: acct477931
[2019-06-11 13:20:09,307] INFO in ocp_report_processor: Cleaning up temporary report files for /var/tmp/masu/acct477931/ocp/test3-ocp-delta
[2019-06-11 13:20:15,609] INFO in ocp_report_processor: Removing /var/tmp/masu/acct477931/ocp/test3-ocp-delta/20190601-20190701_5c979ea3-67de-411a-9a2c-89f735ead5c9_openshift_usage_report.1.csv, completed processing on date 2019-06-11 13:19:40.353455+00:00
[2019-06-11 13:20:15,609] INFO in ocp_report_processor: Removing /var/tmp/masu/acct477931/ocp/test3-ocp-delta/20190601-20190701_5c979ea3-67de-411a-9a2c-89f735ead5c9_openshift_usage_report.0.csv, completed processing on date 2019-06-11 13:19:36.360707+00:00
[...]

PR #463, opened 2019-06-12
Catch masu-listener line item processing exceptions

dccurtis:
Addresses #462 

The asyncio method `run_in_executor` should be ran in a try/catch block so that any exceptions that are thrown unblocks the event loop.

**Testing**
1.  Make a model change to simulate the issue that was seen in issue #462.  In this case removing the `setup_complete` from the`Provider` model was easy to do.
2. Ingest OCP payload and observe the exception occurring.  Verify that the payload could be sent and processed two times. (before this change the second upload of the payload would never be processed since the first was blocking the event loop.)

**Test Results**
1.
Koku Change
```
diff --git a/koku/api/provider/models.py b/koku/api/provider/models.py
index 390aebf..f18171a 100644
--- a/koku/api/provider/models.py
+++ b/koku/api/provider/models.py
@@ -83,7 +83,6 @@ class Provider(models.Model):
                                  on_delete=models.PROTECT)
     created_by = models.ForeignKey('User', null=True,
                                    on_delete=models.SET_NULL)
-    setup_complete = models.BooleanField(default=False)


 class ProviderStatus(models.Model):
```
2.
Upload nice generated payload to upload service two times.
```
masu_listener    | [2019-06-12 02:16:50,921] INFO in kafka_msg_handler: Listener started.  Waiting for messages...
masu_listener    | [2019-06-12 02:16:55,804] INFO in kafka_msg_handler: Successfully extracted OCP for my-ocp-cluster-1/20190601-20190701
masu_listener    | [2019-06-12 02:16:55,820] INFO in kafka_msg_handler: Validating message: b'{"request_id": "52df9f748eabcfea", "validation": "success"}'
masu_listener    | [2019-06-12 02:16:56,765] INFO in kafka_msg_handler: Processing report for account {'authentication': 'my-ocp-cluster-1', 'customer_name': 'acct10001', 'billing_source': '', 'provider_type': 'OCP', 'schema_name': 'acct10001', 'provider_uuid': 'f60fbef4-7e0f-4d28-bb41-86be5607e752'}
masu_listener    | [2019-06-12 02:16:57,137] ERROR in kafka_msg_handler: Thread exception: 'api_provider' object has no attribute 'setup_complete'
masu_listener    | [2019-06-12 02:18:29,497] INFO in kafka_msg_handler: Successfully extracted OCP for my-ocp-cluster-1/20190601-20190701
masu_listener    | [2019-06-12 02:18:29,563] INFO in kafka_msg_handler: Validating message: b'{"request_id": "52df9f748eabcfea", "validation": "success"}'
masu_listener    | [2019-06-12 02:18:30,514] INFO in kafka_msg_handler: Processing report for account {'authentication': 'my-ocp-cluster-1', 'customer_name': 'acct10001', 'billing_source': '', 'provider_type': 'OCP', 'schema_name': 'acct10001', 'provider_uuid': 'f60fbef4-7e0f-4d28-bb41-86be5607e752'}
masu_listener    | [2019-06-12 02:18:30,803] ERROR in kafka_msg_handler: Thread exception: 'api_provider' object has no attribute 'setup_complete'
```


masu/external/kafka_msg_handler.py
```diff
@@ -287,6 +287,7 @@ def process_report(report):
         LOG.info('Summarization celery uuid: %s', str(async_id))
 
 
+# pylint: disable=broad-except
```
blentz: Is this pylint disable truly necessary? Can we narrow the scope of the except block any?

If it is necessary, please include a comment about why we need such a broad except.

dccurtis: Yes that's a good idea.  I don't like the `except Exception` either.  In this case though since it can potentially block the event loop I wanted to be extra loose with the exception catching.  This block can have errors from sources like database, AWS boto APIs, Masu processing exceptions, etc.  Since I *really* don't want the service to be completely locked on any error I thought just keeping the broad `Exception` was ok here.  But I'll add a comment explaining this.

dccurtis:
Hitting a bad intermittent failure in `test_populate_ocp_on_aws_cost_daily_summary`.  Will fix in a separate PR, but will probably continue to hit re-test for now to get this in :)

PR #479, opened 2019-07-18
Fix case where summarization blows up on missing data

adberglund:
## Summary
This makes a WHERE clause expecting bill_ids to be fully optional so that the SQL does not blow up with a syntax error looking for `bill_id IN ()`. This shouldn't be a case that happens in the normal business flow, but when running the report data summary API it triggers AWS providers with no data and blows up.


blentz:
LGTM. Once the coverage checks are passing, I can :+1:

adberglund:
This covered an issue that only occurred when running the report_data API for all providers. It would trigger a run against an AWS provider and if that provider existed but had no data, it would crash. This fixes that so we no longer crash in this case.

Issue #473, opened 2019-07-01
Report files are incorrectly being re-processed


dccurtis:
https://github.com/project-koku/masu/pull/472 removes the check for an exact match of files processed for the manifest before summary is allowed to occur.  Some additional context was also included in that PR around how celery task ordering has changed.

In our production environment we have seen occurrences of OCP and AWS processing more files than are present in the manifest.

One example of this for OCP was seen on the Masu-listener.  (It's not clear if this is the only cause of this problem but it's the only lead we have so far)
```
[2019-07-01 13:59:49,333] INFO in ocp_report_processor: Initialized report processor for file: /var/tmp/masu/acct6089719/ocp/test0-ocp-alpha/20190601-20190701_27d69a85-9252-496f-97a2-1eaf2876e2cf_openshift_usage_report.0.csv and schema: acct6089719

[2019-07-01 13:59:53,294] INFO in ocp_report_processor: Initialized report processor for file: /var/tmp/masu/acct6089719/ocp/test0-ocp-alpha/20190601-20190701_27d69a85-9252-496f-97a2-1eaf2876e2cf_openshift_usage_report.0.csv and schema: acct6089719
```

The provider stats looks like this when this problem occurs:
```
      },
            "stats": {
                "2019-07-01": [
                    {
                        "assembly_id": "c114f2fc-7a67-4ac8-ba59-7f46d23afd66",
                        "billing_period_start": "2019-07-01",
                        "files_processed": "3/2",
                        "last_process_start_date": "2019-07-01 10:42:54",
                        "last_process_complete_date": "2019-07-01 10:42:55",
                        "summary_data_creation_datetime": null,
                        "summary_data_updated_datetime": null,
                        "derived_cost_datetime": "2019-07-01 10:43:32"
                    }
                ],
                "2019-06-01": [
                    {
                        "assembly_id": "6dd60862-95ff-4052-892d-66e3f3543534",
                        "billing_period_start": "2019-06-01",
                        "files_processed": "4/2",
                        "last_process_start_date": "2019-07-01 10:42:37",
                        "last_process_complete_date": "2019-07-01 10:42:46",
                        "summary_data_creation_datetime": null,
                        "summary_data_updated_datetime": null,
                        "derived_cost_datetime": "2019-07-01 10:43:27"
                    }
                ]
            },
            "infrastructure": "Unknown"
        }
```

adberglund:
I believe #490 covers this same issue.

lcouzens:
Verifed

PR #492, opened 2019-07-31
Bugfix duplicate file processing

adberglund:
## Summary 
This changes the logic of when we download and process files for a particular manifest. It should cover the following four cases.

1. A brand new manifest -- download and process.
2. A manifest we have seen before and is completely done processing -- skip.
3. A manifest we have seen before and is currently processing -- skip.
4. A manifest we have seen before, is not finished, and has not processed for some time -- redownload and reprocesses as if it is new.

Case 3 covers a bug we currently have going where we are duplicating data from an AWS provider with multiple files by launching additional download/processing tasks before the first task is able to finish. 

Case 4 covers the case where something goes wrong or our worker spins down mid-processing, we should then go back and re-process. 

This might also cover https://github.com/project-koku/masu/issues/473, but I would like to verify that first.

## Testing
1. Locally downloaded and processed our AWS cost and usage report. Everything worked.
2. Altered manifest to have `num_processed_files` be less than `num_total_files` and re-ran the download API endpoint. We did not download or process as it recognized we were mid processing.
3. Then set the `reporting_common_costusagereportstatus.last_completed_datetime` to be over an hour in the past and reran the download API endpoint. We downloaded and reprocessed the manifest. Line item table had all expected data and no duplicates.
4. After the successful processing was complete I reran the download API endpoint and we did not download and reprocess.


masu/external/downloader/report_downloader_base.py
```diff
@@ -47,6 +49,61 @@ def __init__(self, download_path=None, **kwargs):
...
+                num_total_files = manifest.num_total_files
+                if num_processed_files < num_total_files:
+                    completed_datetime = manifest_accessor.get_last_report_completed_datetime(
+                        manifest_id
+                    )
+                    if completed_datetime and completed_datetime < last_completed_cutoff:
+                        # It has been more than an hour since we processed a file
+                        # and we didn't finish processing. We should download
+                        # and reprocess.
+                        manifest_accessor.reset_manifest(manifest_id)
```
blentz: Are you worried about this becoming an infinite loop?

If manifest processing gets to a point where we never finish processing, but always reset and retry, that seems like it could result in an infinite loop.

adberglund: We can certainly boost the `last_completed_cutoff` which would make it take longer before we re-process something. Right now the criteria for this to trigger is that we haven't finished processing any file in the manifest in the last hour. It takes much less than an hour to process a single file in a manifest, so as long as processing is proceeding normally we shouldn't hit this. It really only covers the case where something went wrong (e.g. we turned the worker off when we released new code) and processing never was able to finish.

blentz: I agree that it's a function of processing time and we're currently well within tolerances.

Because there's a limitation here, I think we ought to monitor for it so that we know when we're approaching that limit. If it's possible to check for this in the code, that's great. But, I think it's reasonable to monitor processing times with Prometheus. (Let's create a new issue to make sure that monitoring gets done, maybe?)

